Issue #1, opened 2019-09-05
Develop of Robust UTC Model in PSim


kylekrol:
Currently, `psim` only supports a trivial model of time. At the start of every run, `truth.mission_time` defaults to zero and simply increments forward by `truth.dt` at every tilmestep. This simple model of time, however, is insufficient to support accurate conversions of vectors and coordinates from ECI to ECEF and vice versa.

To perform these conversions, we will need to leverage the MATLAB functions documented at the following link - I think the `ecef2lla`, `eci2lla`, `lla2ecef` and `lls2eci` functions may be the way to go: https://www.mathworks.com/help/aerotbx/axes-transformations.html?s_tid=CRUX_lftnav

We will also need the ECEF to ECI DCM to convert vectors between the coordinate systems. This should be possible given the `dcmeci2ecef` function at the following link:
https://www.mathworks.com/help/aerotbx/ug/dcmeci2ecef.html

The last thing we need to keep in mind is to ensure we have a way from converting from our UTC model to the proper GPS time that the satellite would see. This will be necessary both for verifying flight code in `sim` and for SHITL testing.

What "time" exactly is passed to the flight computer from Piksi, how are we representing it on the flight computer (we'll need this for the sim later), and do we have a proposed conversion from UTC to GPS time?

@stewartaslan @nhz2 @tanishqaggarwal

nhz2:
The Piksi passes GPS time to the flight computer. GPS time, unlike UTC doesn't have random leap seconds which is nice. For all of the models, time is a double or float in seconds since the start of some GPS week( currently set as week 2045 around march 17th 2019). UTC isn't very useful for precise physical modeling because of the leap seconds.

Technically to get the ECEF to ECI DCM, UT1, which is a measure of how much the earth has rotated, should be used instead of GPS or UTC time, but that level of precision is hopefully not needed, especially if we are just using a J8 gravity model, where the earth is assumed to be rotationally symmetric anyways. 

With the J8 model, the only important effect of the earth's rotation is in converting from velocities in an inertial frame to velocities in ECEF(which is what Piksi returns) because the of the  ω×r part of the velocity. For this ω is usually assumed to be constant.

tanishqaggarwal:
In Flight Software, the time is managed as follows:
- Stored in terms of milliseconds since OS bootup (this "system time" is provided to us by ChibiOS)
[...]

PR #3, opened 2019-09-08
added time convertion utilities

nhz2:
#1 Added some utility functions to convert between matlab datetime objects and the time used on the flight computer. In matlab, datetime objects take care of leap seconds and have ns precision if TimeZone is 'UTCLeapSeconds'.

### DIFF
--- a/MATLAB/config.m
+++ b/MATLAB/config.m
@@ -28,6 +28,8 @@ Global variables treated as outputs:
 const.mu = 3.986e14;      % Earth's gravitational constant (m^3/s^2)
 const.dt = int64(0.1e9);  % Simulation timestep            (ns)
 const.INITGPS_WN= 2045;   % initial gps week number, epoch for time.
+const.MAXWHEELRATE= 677.0; % Max wheel rate in rad/s
+const.MAXWHEELRAMP= 304.5; % Max wheel ramp in rad/s/s
 
 truth.mission_time = int64(0);  % Mission time (ns)
 % ^^ Should always initially be zero
--- a/MATLAB/config.m
+++ b/MATLAB/config.m
@@ -1,6 +1,7 @@
 function config
 
-global truth const
+global truth 
+global const
 
 %{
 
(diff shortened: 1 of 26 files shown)


MATLAB/config.m
```diff
@@ -1,6 +1,7 @@
 function config
 
-global truth const
+global truth 
+global const
```
kylekrol, line 4: Is it actually a problem if the globals aren't declared on separate lines?

MATLAB/environmental_models/helper_functions/wrldmagm_noerror.m
```diff
@@ -0,0 +1,249 @@
+function [XYZ, H, DEC, DIP, F] = wrldmagm_noerror(height, lat, lon, dyear, varargin )
+%  WRLDMAGM Use World Magnetic Model.
+%   [XYZ, H, DEC, DIP, F] = WRLDMAGM(HEIGHT, LAT, LON, DYEAR ) calculates the 
+%   Earth's magnetic field at a specific location and time using the World 
+%   Magnetic Model (WMM).  The default WMM is WMM-2015 which is valid from
+%   January 1, 2015 until December 31, 2019.
```
kylekrol, line 6: Are we going to be updating this in the future to work in 2020+?

MATLAB/utl/datetime2time.m
```diff
@@ -0,0 +1,8 @@
+function time = datetime2time(in_datetime,init_GPS_week_number)
```
kylekrol: Don't want to be super picky, but could we prefix all of the utility functions with`utl_`and maybe environmental models with `env_`? I was just following something like this to make it easier to read once all of the directories get included. Also could help avoid naming collisions.
If this is a bunch of work don't worry about it tho.

kylekrol (commented):
Okay, couple things really quick:
 * Do you want to substitute in the gravity model into the orbit propagator in truth_update so the sim runs off of that?
 * If I understand the new added functions correctly, we have a way to track time? Would you be comfortable adjusting what we have currently (the ns mission time counter) or is the current time tracking system sufficient?
 * Couple other comments down there in the code.

MATLAB/config.m
nhz2: No I don't think so.

MATLAB/environmental_models/helper_functions/wrldmagm_noerror.m
nhz2: Yes, WMM-2020 will be released December 10, 2019. I'm not sure how long Matlab will take to update their stuff though. For now I just removed the error for inputing dates 2020+, on line 93.

stewartaslan (approved):
Still need to increase performance by reducing unnecessary rotations and interpolations. It probably doesn't make sense to do any linear interpolation with ODE2.

PR #6, opened 2019-10-05
Dev/usb console

tanishqaggarwal:
Here's a big boi for you as well @kkrol27 

This provides the desktop-side system for reading and writing to state fields on any Teensy that supports the State Field System (such as the flight controller.) I've also added Travis CI to the repo.

The remaining work (that will be committed to another branch) is to actually interface this with the MATLAB sim.

**High level structure**: `SimulationRun` is meant to represent any run of the simulation, and has the following subobjects:
   - `Simulation`, which runs the MATLAB computations and adjusts state fields accordingly
   - `StateSession`s, which manage the serial connections with Teensies, and have the following subobjects:
       - `Datastore`: A threaded object that stores a record of all telemetry exchanged with the Teensy.
       - `Logger`: A threaded object that stores a record of all debugging output received from the Teensy.
   - `StateCmdPrompt`, which is a user-facing CLI for manually managing state.

The `SimulationRun` sets up the `Simulation`, any required `StateSession`s and their associated `Datastore`s and `Logger`s. It fires up the `Simulation` and the `StateCmdPrompt` and thus begins interaction with the connected Teensies. Upon ending of the simulation or exiting from `StateCmdPrompt`, the `SimulationRun` cleans up all the objects and threads that it created.


python/usb_console/configs/fc_and_adcs.json
```diff
@@ -0,0 +1,15 @@
...
+            "port": "/dev/cu.usbmodem47349801",
+            "stop_if_unconnected": true
+        },
+        {
+            "name": "ADCSController",
+            "port": "/dev/cu.usbmodem",
+            "stop_if_unconnected": true
+        }
+    ]
+}
```
kylekrol: Are these configs generated or no? Just curious because everyone's port(s) will most likely be different.

python/usb_console/configs/fc_only.json
```diff
@@ -0,0 +1,10 @@
+{
+    "seed" : 2,
+    "devices" : [
+        {
+            "name": "FlightController",
+            "port": "/dev/cu.usbmodem47349801",
```
kylekrol: Same as above ^^

python/usb_console/pan_logo.txt
```diff
@@ -0,0 +1,12 @@
...
+  | |__) | / _ \      |   \ | |                               
+  |  ___/ / ___ \     | |\ \| |                              
+ _| |_  _/ /   \ \_  _| |_\   |_                             
+|_____||____| |____||_____|\____|                              
+                                                                
+Pathfinder for Autonomous Navigation                            
+Space Systems Design Studio, Cornell Unversity       
+Console Software | Version 1.0
+
+Type "help" for a list of commands.
```
kylekrol, line 12: Lol I like it

python/usb_console/run_simulation.py
```diff
@@ -0,0 +1,121 @@
...
+                device_logger.start()
+            elif device["stop_if_unconnected"]:
+                print("Error: exiting due to a required serial port being disconnected.")
+                self.stop_all()
+                return
+
+        # Set up MATLAB simulation
+        self.sim = Simulation(self.devices, self.random_seed)
+        self.sim_thread = threading.Thread(name="Python-MATLAB Simulation Interface", target=self.sim.run, args=[2])
+        self.sim_thread.start()
```
kylekrol: We may want to move the thread spawning into the simulation object. Just because when we actually sim two satellites we will have two MATLAB instances running in parallel which will give us much better performance.

python/usb_console/run_simulation.py
```diff
@@ -0,0 +1,121 @@
...
+
+from argparse import ArgumentParser
+from state_session import StateSession
+from cmdprompt import StateCmdPrompt
+from data_containers import Logger, Datastore
+from simulation import Simulation
+import json, sys, os, threading, time
+
+class SimulationRun(object):
+    def __init__(self, random_seed, data_dir, device_data):
```
nhz2: What types are these inputs supposed to be?

python/usb_console/run_simulation.py
nhz2: We're probably going to be bottle necked by communication not computation.

python/usb_console/pan_logo.txt
tanishqaggarwal: Haha thanks :)

python/usb_console/configs/fc_and_adcs.json
tanishqaggarwal: These are not autogenerated, but I put them here so that they're available as templates. They're easy to reconfigure.

python/usb_console/run_simulation.py
tanishqaggarwal: Got it, I'll address this.

python/usb_console/run_simulation.py
tanishqaggarwal: random_seed is an int, data_dir is a string, device_data is a JSON array (see the config files.)

python/usb_console/cmdprompt.py
```diff
@@ -49,7 +49,7 @@ def do_sc(self, args):
         args = args.split()
 
         if len(args) < 1:
-            print('Need to specify a state field to read')
+            print('Need to specify a device to command')
```
tanishqaggarwal, line 52: Generally with PRs, it's better to let the author know about changes and make them handle it--that way the reviewers aren't tainted by having changes on the commit. I might have missed this if I didn't `git pull`. Thanks for doing this though! @nhz2

python/usb_console/run_simulation.py
kylekrol: Oh really? Okay. Wasn't aware we had the sim moving that much faster. We can discuss com time vs comp time in slack or something as well.

python/usb_console/run_simulation.py
kylekrol: hahahahaha python and types....

kylekrol (dismissed):

kylekrol (dismissed):

kylekrol (approved):

PR #7, opened 2019-10-06
Attitude controllers

nhz2:
Work in progress sensor models and attitude controller. Also I changed main, so it doesn't make plots any more, instead, to plot the trajectory, run main, then run plot_trajectory(truth_trajectory).


MATLAB/main.m
```diff
@@ -2,73 +2,49 @@
...
+    truth=orbit_attitude_update_ode2(truth,actuators,double(const.dt) * 1e-9);
+    %update time
+    truth.mission_time= truth.mission_time+const.dt;
+    truth.time= double(truth.mission_time)*1E-9;
+    %update sensor state (for example biases)
+    sensor_state_update(sensor_state,truth,double(const.dt) * 1e-9);
+    %simulate flight computer
+    [computer_state,actuator_commands]=update_FC_state(computer_state,sensor_readings);
+    %command actuators
+    actuators=actuator_command(actuator_commands,truth);
```
kylekrol: Looks great

kylekrol (commented):
I think the overall structure looks pretty good. Not gonna be too terribly picky given so many things are going to be changing so quickly. I'm good with it if @stewartaslan is as well.

Won't approve it for now, but if no one else does and you want to keep moving @nhz2, lmk.

stewartaslan:
Before we merge this PR I'd like to improve performance a little bit, since it no longer runs in real time. Once we improve the performance to something like 2x or 10x real time let's merge this stuff.

stewartaslan (approved):
Sim now runs at 10-20x real time, since we replaced the call to planetEphemeris with the simple Keplerian model of the Earth's orbit around the sun. The unit test demonstrates that this simplification is always within 0.1 degrees of the truth.

PR #11, opened 2019-10-07
MATLAB console integration

tanishqaggarwal:
This PR marks successful integration between the Python script and the MATLAB state console.

Currently I'm not transacting any variables between the MATLAB simulation and the Python system; this will change once we create controllers that can run on Flight Software. However, the simulation runs stably in the background.

One thing that's really scary is that state transactions between the Python system and a Teensy/native binary take *~100 ms* to complete. We need a 1000x order reduction in this process and I'll be looking aggressively at ways we can do this. My suspicion is that inter-process calls aren't fast enough.

Another thing that doesn't work is the logging of `truth_trajectory` to `simulation_data.txt`. Not sure why.

Summary of changes:
- State console profiles the timing of each read-write transaction
- MATLAB simulation is stable within Python
   - Truth trajectory is logged to a text file
- Directions for arriving at stable setup
- Logging is a little more stable
- Unit testing for logging

**Edit:** Disregard my comment above about speed. The 100 ms delay was due to a delay I'd put in the Flight-Software side to mimic the duration of a control cycle. The actual round-trip time for a single state field is ~0.5 ms (worst case ~2 ms, I've noticed.) Still not great, but if we batch the sends, it'll be fine.


python/usb_console/simulation.py
```diff
@@ -18,11 +23,14 @@ def __init__(self,devices,seed):
...
+        self.log = ""
 
     def start(self, duration):
         self.sim_thread = threading.Thread(name="Python-MATLAB Simulation Interface",
                                            target=self.run,
                                            args=[duration])
+        self.running = True
         self.sim_thread.start()
 
     def run(self,duration):
```
tanishqaggarwal: This function essentially does the same thing as main.m in the MATLAB/ folder. It has to, in order to run the simulation.

stewartaslan (approved):
Everything looks good. Last time I ran it there was no evidence the sim was running in the background, but that might have been fixed by now.

PR #12, opened 2019-10-08
Dev/faster earth attitude model

nhz2:
20x faster earth rotation model ported over from Simulink with 0.1 degree accuracy over 5 years.


kylekrol (commented):
All looks good to me. @stewartaslan you wanna take a look over it as well?

stewartaslan (approved):
Looks good to me.

PR #18, opened 2019-10-11
Desaturation, B-dot controllers, and simple adcs state machine.

nhz2:


stewartaslan (dismissed):
This looks great to me! Can you make a flowchart diagram or something listing all the states in the state machine? I think it would be good to have something like that in the documentation so that we can keep track of the states and make sure they align with the states Tanishq is using in flight software.

kylekrol:
Looks great!

kylekrol (approved):

Issue #23, opened 2019-10-15
actuator_command


nhz2:
Modify the state of one satellite, given its actuator commands.

tanishqaggarwal:
Closed? @nhz2

tanishqaggarwal:
Very old

Issue #22, opened 2019-10-15
sensor_reading


nhz2:
Outputs the sensor readings of one satellite.

nhz2:
To test the control algorithms its important to see if they still work with sensor noise and bias. For attitude control this is mainly gyro bias and noise and magnetometer bias and noise.

Here is some data recorded outside from the two magnetometers. The magnetometers will probably have additional noise and bias from the wheels, magnetorquers, and docking magnets.

The sensor noise models don't have to be super realistic, they just have to be worse than the real worst case noise/bias.

nhz2:
Also the sunsensor model and when the sunsensors can't see the sun.

kylekrol:
In general when there isn't sufficient lighting to detect the sun the ADCS will just return a flag that indicates the sun isn't in view. It only includes a photodiode in the sun vector calculation if the observed voltage is greater than 50% of the maximum expected intensity.

Would this sort of model work in PSim?

nhz2:
Yeah, just modify `sensor_reading.m`

Right now sun sensor calculations are done starting at line 30

I am putting the sun sensor normals in `my_satellite_state.sensors.sunsensor_real_normals` and `my_satellite_state.sensors.sunsensor_measured_normals`: These are size (3,20) matrixes initialized in `initialize_main_state.m`. Should these should have units of volts instead of unitless? Right now they are just the ideal unit vectors. To get the true sun vector, you can call `get_truth('sat2sun body',my_satellite_state.dynamics)`, and implement that functionality in `get_truth.m`, or for now, since `get_truth.m` doesn't exist yet, you can just use the quaternions and `env_sun_vector.m`.

kylekrol:
@nhz2 Do we want to keep this open or close it in favor of making multiple, more specific issues - like #79?

tanishqaggarwal:
Very old

Issue #21, opened 2019-10-15
main_state_update


nhz2:
Updates the main state over one delta time step.

tanishqaggarwal:
Closed? @nhz2

Issue #20, opened 2019-10-15
initialize_main_state


nhz2:
Sample an initial main state given a seed and condition
To set the seed based on current time, use seed = 'shuffle'

tanishqaggarwal:
Very old

PR #24, opened 2019-10-15
Sim restructuring

nhz2:
#23 #22 #21 #20 #19 

A lot of new stuff in the readme, and templates for the functions python will call.

### DIFF
--- a/MATLAB/README.md
+++ b/MATLAB/README.md
@@ -3,26 +3,243 @@
 
 This directory will contain all MATLAB code housed in this repository.
 
-The MATLAB portion of the simulation should house a sequence of scripts that
-operate on variables stored in the global namespace. For example, we could have
-`truth.r` store the truth position of the satellite in ECI and `truth.v` store
-it's truth velocity in ECI. `truth.rwa_axl` can store reaction wheel angular
-acceleration, `truth.rwa_w` can store the reaction wheels angular speeds, et
-cetera.
+The MATLAB portion of psim has two major components.
 
-If we have a sequence of scripts that can sim a single satellite, we can then
-run a second inpedendant sequence of scripts via python (see the `/python`
-folder) in parallel to sim a full mission. Yes, we would need to pass a little
-bit of information between the MATLAB instances to perform rendezvous
-calculations, but this can easily be done through python and given only a small
-amount of data will be passed, the overhead will be tiny.
+1. A simulator to model the physics of pan.
+2. A matlab prototype version of gnc related flight software.
 
-Also, once a full simulation is complete, we will need to add scripts that run
-conditionally depending on whether or not the satellite is the leader, follower,
-et cetera. We will be able to slowly replace MATLAB code with CXX code to test
+
+The simulator has four main functions that preform computations with the main state.
+    initialize_main_state: Constructs the main state given a seed, and situation.
+    sensor_reading: Outputs the sensor readings of one satellite.
+    main_state_update: Updates the main state over one delta time step.
+    actuator_command: Modifies the state of one satellite, given its actuator commands.
+
+The matlab prototype of gnc flight software has two main functions.
+    initialize_computer_states: Constructs both leader and follower computer states.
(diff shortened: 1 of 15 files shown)


MATLAB/main.m
```diff
@@ -1,74 +1,49 @@
...
-    tic;
-    sensor_state_update(sensor_state,truth,dt);
-    t_sensor_state = toc;
-    %simulate flight computer
-    tic;
-    [computer_state,actuator_commands] = update_FC_state(computer_state,sensor_readings);
-    t_fc = toc;
+    main_state = main_state_update(main_state);
+    %simulate flight computers
+    [computer_state_follower,actuator_commands_follower] = update_FC_state(computer_state_follower,sensor_readings_follower);
```
tanishqaggarwal, line 34: Could you clarify, again, what the update_FC_state function does? How should this play with sim?

MATLAB/README.md
```diff
@@ -3,60 +3,295 @@
...
+
+get_truth('rate body eci eci',main_state.follower.dynamics)
+returns the angular rate of the transform from eci to body in the eci frame.
+
+get_truth('velocity eci',main_state.follower.dynamics)
+returns the velocity of the center of mass of the sat relative to earth in the eci frame.
+
+get_truth('velocity ecef',main_state.follower.dynamics)
+velocity measurements take into account additional velocity from rate cross radius.
+returns the velocity of the center of mass of the sat relative to earth in the ecef frame.
```
kylekrol, line 90: I think this is great and super clean way of doing it. For some reason thought it'd be worse than this on the phone. Even if we can't call the functions directly from python or something we can just dumped them over all at one time.

MATLAB/README.md
```diff
@@ -3,60 +3,295 @@
...
+
+get_truth has full support for the following frames:
+ * body: The +x face has the antennas, the -z face has the docking magnets.
+ * eci: Earth Centered Inertial, an inertial frame, z axis is close to(but not exactly) the north pole.
+ * ecef: Earth Centered Earth Fixed, rotates with earth, z axis is the north pole.
+ * lvlh: Local Vertical Local Horizontal, x axis is position from earth to sat, z axis is orbit normal
+ * vbn: Velocity Binormal Normal, x axis is sats eci velocity, z axis is orbit normal.
```
kylekrol, line 97: Big thumbs up.

MATLAB/main.m
kkrol27: Correct me if I'm wrong @nhz2, but this should be an actual call to the teensy during a HITL. But for the MATLAB standalone version we'll have the cut down state machine Nathan has going (mainly just handles ADCS states).

kylekrol (approved):
All in all I think it looks awesome - great job!

MATLAB/main.m
tanishqaggarwal: Since the Teensy can't handle all state values yet, how can we make this function handle partial state updates from the Teensy?

kylekrol: Correct me if I'm wrong @nhz2, but this should be an actual call to the teensy during a HITL. But for the MATLAB standalone version we'll have the cut down state machine Nathan has going (mainly just handles ADCS states).

@tanishqaggarwal  This may be a conversation to have some other place, but what can the flight computer handle at his time?

MATLAB/README.md
nhz2: Most of the matlab functions should be callable from python. The matlab engine automatically converts python strings to matlab strings, and python dictionaries to matlab structures. [python matlab type conversions](https://www.mathworks.com/help/matlab/matlab_external/pass-data-to-matlab-from-python.html)

stewartaslan (approved):
Looks great, fantastic work! Very thorough documentation as well.

MATLAB/main.m
tanishqaggarwal: There's no controllers yet written for the flight controller, so there's no state reported yet by the flight controller.

PR #30, opened 2019-10-16
Initial draft of QuakeSession object

tanishqaggarwal:
This is the core object within the future Ground Software. This object will be responsible for checking downlinks and sending uplinks to one Quake radio via the Iridium satellite constellation.

Putting this out there so that the design for it is on the back of our minds. @fatimayousuf Could you start thinking about how to integrate your email script into this? For now, I think we can pretend that the communication protocol over Quake is just JSON--we can make the protocol much more efficient later on.

### DIFF
--- a/python/usb_console/configs/README.md
+++ b/python/usb_console/configs/README.md
@@ -1,0 +1,13 @@
+# USB Console Configuration
+
+## Items
+- `seed` : Randomization seed for simulation.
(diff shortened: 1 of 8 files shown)


stewartaslan (approved):
Looks good to me for Quake testing. Just to be clear, this is not part of psim, but rather a test that, similar to psim, can be run from the USB console.

PR #31, opened 2019-10-16
Update python simulation with latest MATLAB code

tanishqaggarwal:
@nhz2 Some path changes were needed to config.m

Simulation data logging to file works properly now, as well.

### DIFF
--- a/python/usb_console/simulation.py
+++ b/python/usb_console/simulation.py
@@ -59,11 +59,8 @@ class Simulation(object):
 
         self.eng.config(nargout=0)
         self.eng.eval("global const", nargout=0)
-        self.eng.eval("global main_state_trajectory", nargout=0)
-        self.eng.eval("global computer_state_follower_trajectory", nargout=0)
-        self.eng.eval("global computer_state_leader_trajectory", nargout=0)
 
-        self.main_state = self.eng.initialize_main_state(1,'not_detumbled', nargout=1)
+        self.main_state = self.eng.initialize_main_state(self.seed, 'not_detumbled', nargout=1)
         self.computer_state_follower, self.computer_state_leader = self.eng.initialize_computer_states('not_detumbled', nargout=2)
         self.main_state_trajectory = []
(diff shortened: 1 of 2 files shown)


python/usb_console/simulation.py
```diff
@@ -53,18 +53,21 @@ def configure_sim(self):
...
-        self.eng.addpath(path4, nargout=0)
+        self.eng.addpath(path, nargout=0)
 
-        self.eng.eval("global const", nargout=0)
         self.eng.config(nargout=0)
+        self.eng.eval("global const", nargout=0)
+        self.eng.eval("global main_state_trajectory", nargout=0)
+        self.eng.eval("global computer_state_follower_trajectory", nargout=0)
+        self.eng.eval("global computer_state_leader_trajectory", nargout=0)
+
```
nhz2: You don't need these in python, they are just lists of the states for plotting at the end of the simulation.

python/usb_console/simulation.py
```diff
@@ -53,18 +53,21 @@ def configure_sim(self):
...
+        self.main_state = self.eng.initialize_main_state(1,'not_detumbled', nargout=1)
```
nhz2: instead of 1 this should be self.seed

python/usb_console/simulation.py
```diff
@@ -77,17 +80,40 @@ def run(self):
...
-        start_time = time.time()
-        while self.sim_time < self.sim_duration and self.running:
-            # Update sensor model and begin update of truth model in the background
-
-            # Send sensor data to Flight Controller, and collect actuator outputs
 
-            # Update actuators that are not updated by flight controller
+        dt = self.eng.workspace['const']['dt'] * 1E-9
+        num_steps = int(self.sim_duration/dt)
+        sample_rate = 10.0 / dt # Sample once every ten seconds
```
nhz2: sample_rate should be an int

nhz2 (approved):

PR #33, opened 2019-10-17
Add Ideal Sun Sensor Model and MATLAB Sun Vector Determination Algorithm

kylekrol:
Relates to #22.

The full algorithm, as will exists on the ADCS computer, has been implemented. There are now two sets of configuration data. First, you have `sensors.sunsensor_*_normals` which is a matrix on unit vectors representing the normal vectors of each sun sensor in the body frame of the spacecraft. Second, there is the `sensors.sunsensor_*_voltage_maximums` which is vector containing a voltage normalization constant to place each individual diodes measurement with the range `[0, 1]`.

From there we rely on a single least squares problem of the following form:

equation

Where a given normal vector and it's corresponding voltage measurement is only included if the voltage measurement is above a specified threshold. This has two benefits:
 1. We avoid issues with the dot product actually going negative when the sun vector would be pointing behind a given photodiode
 2. We can ignore edge effects when by setting a relatively high voltage threshold (like `0.5 V` which maps to `30` degrees)

Once the lower voltage measurements are removed from the above system, we solve for a sun vector if we have at least 4 rows in the matrix and RHS vector (this threshold of 4 can be played with but cannot drop below 3 for obvious reasons).

The algorithm was verified correctly by by observing multiple iterations with the MATLAB debugger and looking at `norm(sun_vec - sat2sun, 'fro')` at the end of calls to `update_sun_sensors`.

### DIFF
--- a/MATLAB/sensor_reading.m
+++ b/MATLAB/sensor_reading.m
@@ -33,18 +33,12 @@ sensor_readings.magnetometer_body=utl_rotateframe(quat_body_ecef,B_ecef')';
 sat2sun_eci=env_sun_vector(true_state.time);
 sat2sun_body=utl_rotateframe(quat_body_eci,sat2sun_eci')';
 
-%now determine if the satellite is in eclipse with earth.
+%determine if the satellite is in eclipse with earth.
 eclipse = env_eclipse(true_state.position_eci,sat2sun_eci);
-sun_in_dead_zone=(sat2sun_body(3))<-cos(30*pi/180);
 
-%we can't read the sun sensor because the sun is in the dead zone
-if ((~eclipse) && (~sun_in_dead_zone))
-    sensor_readings.sat2sun_body=sat2sun_body;
-    sensor_readings.sun_sensor_true= true;
-else
-    sensor_readings.sat2sun_body= NaN(3,1);
-    sensor_readings.sun_sensor_true= false;
-end
+[sensor_readings.sat2sun_body,...    %unit vector in the body frame
+ sensor_readings.sun_sensor_true,... %true if succesfull
+] = update_sun_sensors(my_satellite_state.sensors, sat2sun_body, eclipse);
 
 %% wheel angular momentum reading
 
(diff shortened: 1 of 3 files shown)


MATLAB/sensor_reading.m
```diff
@@ -58,3 +55,57 @@
...
+
+if (~eclipse)
+
+    % TODO : Revisit voltage measurement generation
+    [~, N] = size(real_n);
+    voltages = zeros(N, 1);
+    for i = 1:N
+        %magnitude of a normal is the inverse of the max voltage reading so
+        %dividing by the frobenius norm gives a voltage measure.
+        voltages(i) = sat2sun' * real_n(:, i) / norm(real_n(:, i), 'fro');
```
nhz2: dividing real_n(:,i) by its norm just normalizes real_n(:,i). So voltages is cos(theta) now.

MATLAB/sensor_reading.m
```diff
@@ -58,3 +55,57 @@
...
+            Y(j) = voltages(i);
+        end
+    end
+    
+    %perform least squares if possible
+    if j >= 4  % TODO : Perhaps play with this parameter
+        A = A(1:j, :);
+        Y = Y(1:j, :);
+        [Q, R] = qr(A);
+        sun_vec = R \ (Q' * Y);
```
nhz2, line 109: Should sun_vec be normalized here? or does the QR automatically do that?

MATLAB/sensor_reading.m
```diff
@@ -58,3 +55,57 @@
...
+        %dividing by the frobenius norm gives a voltage measure.
+        voltages(i) = sat2sun' * real_n(:, i) / norm(real_n(:, i), 'fro');
+    end
+
+    A = zeros(N, 3);
+    Y = zeros(N, 1);
+    j = 0;
+    for i = 1:N
+        %enforce voltage threshold for inclusion in sun vector calculation
+        if voltages(i) > 0.5
```
nhz2, line 99: On the adcs software, does each diode have its own voltage threshold, or is each diodes's voltage normalized to 0 to 1 before the calculation?

kylekrol:
@nhz2 @stewartaslan This should be updated and ready to go. I left a mathematical explanation of the algorithm in the PR description; I think it's relatively straightforward. Also made updates to the README documenting the new state fields and updating the progress on some of the CXX stuff.

nhz2 (approved):

PR #38, opened 2019-10-17
faster mag model

nhz2:


nhz2:
Josh found that the magnetic field model coefficients were loaded from a file every time the magnetic field function was called. Now its only loaded in config, speeding up the sim significantly.

jer296 (approved):

PR #39, opened 2019-10-19
Made the simulation over 10x faster

nhz2:
Using the 'Run and Time' button in Matlab, I found all of the slow parts of the code. First the builtin magnetic field model had a lot of really bad code, like passing a pointer for two different sin and cos functions based on whether degrees or radians were used. The next slow part was the builtin quaternion class. I replaced almost all uses of the quaternion class with just 4x1 arrays, and added more utl functions to do the quaternion math. Also, instead of calling ode2, I'm just doing the integration in `dynamics_update`. Lastly I added my C++ version of geomag, and a wrapper for calling it in Matlab.


include/geomag.hpp
```diff
@@ -0,0 +1,148 @@
...
+The data from WMM2015, is not subject to copyright protection.
+Modifications are:
+  using ITRS coordinates,
+  conversion from nT to T,
+  Using unnormalized coefficents genrated by the python script wmmcofparsing.py
+  using a different spherical harmonics calculation, described in sections 3.2.4 and 3.2.5:
+    Satellite Orbits Models, Methods and Applications,
+      by Oliver Montenbruck and Eberhard Gill 2000
+*/
+#pragma once
```
tanishqaggarwal: We should try to avoid #pragma once. I know it's clean, but it's not standard code practice yet. No telling where it breaks a CI sometime.

That is to say, it might be _pragmatic_, but unfortunately still unsupported in many areas.

MATLAB/dynamics_update.m
```diff
@@ -124,10 +114,12 @@
...
     init_y(14:16)=initial_state.fuel_net_angular_momentum_eci;
     final_time=initial_state.time+delta_time;
-    opts = odeset('RelTol',1E-12,'AbsTol',1e-9);
+    %opts = odeset('RelTol',1E-12,'AbsTol',1e-9);
     %[~,ys]=ode45(@state_dot,[initial_state.time,final_time],init_y,opts);
-    ys=utl_ode2(@state_dot,[initial_state.time,final_time],init_y);
-    final_y= ys(end,:)';
+    %ys=utl_ode2(@state_dot,[initial_state.time,final_time],init_y);
+    %final_y= ys(end,:)';
+    final_y= init_y+state_dot(initial_state.time,init_y)*delta_time;
```
kylekrol: Just wanna make sure we verified that ode1 is sufficiently accurate when compared with our previous integrator.

kylekrol (dismissed):

kylekrol (dismissed):

kylekrol (dismissed):

kylekrol:
Next time I’d look at rebasing instead of merging master into a dev branch. It essentially replays all of the changes you’ve made on top of a different commit. So if you have your local master branch updated, your dev branch checked out, and run ‘git rebase master’ it’ll look like your dev branch was split off from the updated version of master. Lastly you just need to force push to your dev branch.

This gets rid of the pesky merge commits.

tanishqaggarwal:
> Next time I’d look at rebasing instead of merging master into a dev branch. It essentially replays all of the changes you’ve made on top of a different commit. So if you have your local master branch updated, your dev branch checked out, and run ‘git rebase master’ it’ll look like your dev branch was split off from the updated version of master. Lastly you just need to force push to your dev branch.
> 
> This gets rid of the pesky merge commits.

I think you should only do this if the pull request isn't under active review. Otherwise the files that were reviewed get unmarked and it creates a pain for your reviewer. Merging is better if your PR is active.

In this case the PR was pretty long-running anyway. It would have been better to split into multiple PRs; Nathan and I discussed that already.

nhz2:
The last commit was just resolving a conflict with the readme in master, no changes to the actual code.

PR #41, opened 2019-10-20
Adding Quaternion Utility Functions with Unit Tests in C++

### LINKED ISSUE #40: Code Additional Quaternion Utility Functions in C++
See the last couple comments of #13.

As @nhz2 pointed out, it'd be very beneficial to mirror our handling of quaternions in our c++ implementations. Therefore quaternions will be stored as `lin::Vector4f`/`lin::Vector4d` types and will require associated MATLAB utility functions to be ported over to c++.

From Nathan's comment, the functions will be:

    utl_triad(const Vector3f& N_sun, const Vector3f& N_mag, const Vector3f& B_sun, const Vector3f& B_mag, Vector4f& result)
    utl_quat_cross_mult(const Vector4f& q1, const Vector4f& q2, Vector4f& result)
    utl_quat_cross_mult(const Vector4d& q1, const Vector4d& q2, Vector4d& result)
    utl_dcm2quat(const Matrix3x3f, Vector4f& result)
    utl_quat_conj(const Vector4f& q, Vector4f& result)
    utl_quat_conj(const Vector4d& q, Vector4d& result)
    utl_rotateframe(const Vector4f& q, const Vector3f& v, Vector3f& result)
    utl_rotateframe(const Vector4d& q, const Vector3d& v, Vector3d& result)

These functions are currently in Matlab in the MATLAB/utl folder.

kylekrol:
Closes #40.

This is the start of porting over the MATLAB code which will fly not the spacecraft. I've set up a unit testing framework with gtest in the repository that builds with CMake (so hopefully windows people can figure out how to get it up and running).

On Mac/linux, to compile and run the tests simply execute the following commands from the root of the PSim repository:

    mkdir build
    cd build
    cmake ../
    cmake --build .
    ctest --verbose .

The tests are built and run with Travis CI as well.

I will be adding documentation on the structure of the C/CXX code, testing framework, and the MATLAB mex wrappers in another PR (coming soon).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,19 @@
-language: python
-python:
-    - "3.6"
-
-sudo: false
-
-before_install:
-    - cd python
-
-install:
-    - pip install -r requirements.txt
-
-script:
-    - python -m pytest
+matrix:
+    include:
+        - language: python
+          python: 3.6
+          sudo: false
+          before_install:
+              - cd python
+          install:
+              - pip install -r requirements.txt
+          script:
+              - python -m pytest
+        - language: cpp
+          sudo: false
+          script:
+              - mkdir build
+              - cd build
+              - cmake ../
+              - cmake --build .
+              - ctest --verbose .
--- a/include/inl/gnc_utilities.inl
+++ b/include/inl/gnc_utilities.inl
@@ -1,0 +1,138 @@
+//
+// include/gnc_utilities.hpp
+// PSim
+//
+// Contributors:
+//   Kyle Krol  kpk63@cornell.edu
+//
+// Pathfinder for Autonomous Navigation
+// Space Systems Design Studio
+// Cornell Univeristy
+//
+
+#include "../gnc_utilities.hpp"
+
+namespace gnc {
+namespace utl {
+
+template <typename T>
+constexpr void quat_conj(lin::Vector<T, 4> const &q, lin::Vector<T, 4> &res) {
+  res = { -q(0), -q(1), -q(2), q(3) };
+}
+
+template <typename T>
+constexpr void quat_conj(lin::Vector<T, 4> &q) {
+  quat_conj(q, q);
+}
+
+template <typename T>
+constexpr void quat_cross_mult(lin::Vector<T, 4> const &q1, lin::Vector<T, 4> const &q2,
+    lin::Vector<T, 4> &res) {
+  auto const qv1 = lin::ref<3, 1>(q1, 0, 0);
+  auto const qv2 = lin::ref<3, 1>(q2, 0, 0);
+  lin::ref<3, 1>(res, 0, 0) = lin::cross(qv2, qv1) + q1(3) * qv2 + q2(3) * qv1;
+  res(3) = q1(3) * q2(3) - lin::dot(qv1, qv2);
+}
+
+template <typename T>
+constexpr void rotate_frame(lin::Vector<T, 4> const &q, lin::Vector<T, 3> const &v, lin::Vector<T, 3> &res) {
+  auto const qv = lin::ref<3, 1>(q, 0, 0);
+  res = v + lin::cross(
+      (static_cast<T>(2.0) * qv).eval(),
+      (lin::cross(qv, v) - q(3) * v).eval()
+    );
+}
+
+template <typename T>
+constexpr void rotate_frame(lin::Vector<T, 4> const &q, lin::Vector<T, 3> &v) {
+  rotate_frame(q, v, v);
+}
+
+// TODO : Potentially add diagonal references to lin
+template <typename T>
+inline void dcm_to_quat(lin::Matrix<T, 3, 3> const &M, lin::Vector<T, 4> &q) {
+  // Find the maximum among the diagonal elements and trace
+  T max = M(0, 0);
+  T trace = M(0, 0);
+  unsigned int idx = 0;
+  for (unsigned int i = 1; i < 3; i++) {
+    trace += M(i, i);
+    if (M(i, i) > max) {
+      max = M(i, i);
+      idx = i;
+    }
+  }
+  if (trace > max) {
+    max = trace;
+    idx = 3;
+  }
+
+  // Calculate the unnormalized quaternion
+  switch (idx) {
+    // Max was M(0, 0)
+    case 0:
+      q = {
+        static_cast<T>(2.0) * max - trace + static_cast<T>(1.0),
+        M(0, 1) + M(1, 0),
+        M(0, 2) + M(2, 0),
+        M(1, 2) - M(2, 1)
+      };
+      break;
+    // Max was M(1, 1)
+    case 1:
+      q = {
+        M(1, 0) + M(0, 1),
+        static_cast<T>(2.0) * max - trace + static_cast<T>(1.0),
+        M(1, 2) + M(2, 1),
+        M(2, 0) - M(0, 2)
+      };
+      break;
+    // Max was M(2, 2)
+    case 2:
+      q = {
+        M(2, 0) + M(0, 2),
+        M(2, 1) + M(1, 2),
+        static_cast<T>(2.0) * max - trace + static_cast<T>(1.0),
+        M(0, 1) - M(1, 0)
+      };
+      break;
+    // Max was trace(M)
+    default:
+      q = {
+        M(1, 2) - M(2, 1),
+        M(2, 0) - M(0, 2),
+        M(0, 1) - M(1, 0),
+        trace + static_cast<T>(1.0)
+      };
+  }
+
+  // Normalize the quaternion
+  q = q / lin::norm(q);
+}
+
+// TODO : Add more specific tolerancing to this function along with the QR
+// linear system solvers in lin/include/algorithms/*
+template <typename T>
+inline int triad(lin::Vector<T, 3> const &N_sun, lin::Vector<T, 3> const &N_mag,
+    lin::Vector<T, 3> const &B_sun, lin::Vector<T, 3> const &B_mag, lin::Vector<T, 4> &res) {
+  // Ensure that sun and magnetic field vectors aren't parallel or anitparallel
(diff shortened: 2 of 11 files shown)


include/inl/utilities.inl
```diff
@@ -0,0 +1,87 @@
...
+constexpr void quat_conj(lin::Vector<T, 4> &q) {
+  quat_conj(q, q);
+}
+
+template <typename T>
+constexpr void quat_cross_mult(Vector4f const &q1, Vector4f const &q2, Vector4f &res) {
+  auto const qv1 = lin::ref<3, 1>(q1, 0, 0);
+  auto const qv2 = lin::ref<3, 1>(q2, 0, 0);
+  lin::ref<3, 1>(res, 0, 0) = lin::cross(qv2, qv1) + q1(3) * qv2 + q2(3) * qv1;
+  res(3) = q1(3) * q2(4) - lin::dot(qv1, qv2);
```
nhz2: should be `q2(3)` instead of `q2(4)`  Matlab 1 based indexing is the worst.

kylekrol: You are totally right lol

tanishqaggarwal:
When this is complete, can you add a companion PR to make psim a library of FlightSoftware? This will let us access this lin code and GNC built on top of this lin code from within FSW.

include/inl/utilities.inl
```diff
@@ -0,0 +1,87 @@
...
+// Cornell Univeristy
+//
+
+#include "../utilities.hpp"
+
+namespace gnc {
+namespace utl {
+
+template <typename T>
+constexpr void quat_conj(lin::Vector<T, 4> const &q, lin::Vector<T, 4> &res) {
```
tanishqaggarwal: Is there type checking to ensure that T is either a double or a float?

kylekrol:
@tanishqaggarwal  Yep - I can do that for sure!

@nhz2 @tanishqaggarwal @stewartaslan Also, just in general, I would like to add unit testing support for c++ stuff with this PR. Is everyone alright using gtest? Given most of us developing in c++ have Macs, for now I could just assume macports?

**EDIT**: I actually got gtest with a simple cmake implementation and adding gtest as a submodule so no need to be platform specific anymore. I'll try to add the unit testing to our CI infrastructure as well - although it's tricky apparently cuz c++14 isn't supported by default from what I've read.

tanishqaggarwal:
Where have you read that? C++14 has been working thus far on Travis CI without a hitch.

You could've just set up a platformio project and leveraged the existing CI setup on FlightSoftware/CommonSoftware as boilerplate--but I guess you already went ahead with gtest. Sorry for not seeing this earlier.

kylekrol:
No worries. I read it in a blog post or how-to somewhere but it doesn't really matter because they've updated the default compilers now. I also didn't really want platformio to be a dependency of PSim if we don't need it to be lol. That being said, I didn't know how to do unit tests natively in platformio anyway so it would've been a learning curve there as well.

include/inl/gnc_utilities.inl
```diff
@@ -0,0 +1,138 @@
...
+  q = q / lin::norm(q);
+}
+
+// TODO : Add more specific tolerancing to this function along with the QR
+// linear system solvers in lin/include/algorithms/*
+template <typename T>
+inline int triad(lin::Vector<T, 3> const &N_sun, lin::Vector<T, 3> const &N_mag,
+    lin::Vector<T, 3> const &B_sun, lin::Vector<T, 3> const &B_mag, lin::Vector<T, 4> &res) {
+  // Ensure that sun and magnetic field vectors aren't parallel or anitparallel
+  if (abs(lin::dot(N_sun, N_mag)) == static_cast<T>(1.0)) return 1;
```
nhz2, line 119: Just let res get set to NaN if there is an error instead of returning error codes. The ADCS controller will check if the mag, or sun vector are within a few degrees before calling triad anyway. If someone doesn't check the error code, it might go unnoticed, and res would be who knows what.

nhz2 (approved):
Looks good. You may want a ASSER_QUAT_NEAR macro, to compare two attitude quaternions. Check that the quaternions are normalized, and ensure the absolute value of their dot product is close to one. Also, some tests for triad.

.travis.yml
```diff
@@ -1,14 +1,19 @@
...
-
-before_install:
-    - cd python
-
-install:
-    - pip install -r requirements.txt
-
-script:
-    - python -m pytest
+matrix:
```
tanishqaggarwal, line 1: This is sick. Can you explain how this works? I never understood the matrix idea.

include/inl/gnc_utilities.inl
```diff
@@ -0,0 +1,138 @@
...
+// Cornell Univeristy
+//
+
+#include "../gnc_utilities.hpp"
+
+namespace gnc {
+namespace utl {
+
+template <typename T>
+constexpr void quat_conj(lin::Vector<T, 4> const &q, lin::Vector<T, 4> &res) {
```
tanishqaggarwal, line 19: Consider adding static_asserts that check if T is a float or double via `std::is_same<T, float>::value`. Unless lin already does this type checking?

test/utilities_test.cpp
```diff
@@ -0,0 +1,64 @@
...
+#include <gnc_utilities.hpp>
+
+#include <gtest/gtest.h>
+
+#define ASSERT_VEC_NEAR(a, b) ASSERT_NEAR(lin::fro(a - b), 0.0, 1e-4)
+
+using namespace gnc;  // Only b/c this repo pretty much defines all of ::gnc
+
+TEST(UtilitiesTest, QuaternionConjugateTest) {
+  lin::Vector4f q = {2.0f, 1.0f, -2.5f, 1.0f};
```
tanishqaggarwal, line 23: Might want to also test near-zero values to make sure output quaternions don't go NaN? I don't know enough quaternion math yet to know if/how this might happen, but wanted to make the suggestion.

tanishqaggarwal (approved):

include/inl/gnc_utilities.inl
```diff
@@ -0,0 +1,138 @@
...
+  auto const qv2 = lin::ref<3, 1>(q2, 0, 0);
+  lin::ref<3, 1>(res, 0, 0) = lin::cross(qv2, qv1) + q1(3) * qv2 + q2(3) * qv1;
+  res(3) = q1(3) * q2(3) - lin::dot(qv1, qv2);
+}
+
+template <typename T>
+constexpr void rotate_frame(lin::Vector<T, 4> const &q, lin::Vector<T, 3> const &v, lin::Vector<T, 3> &res) {
+  auto const qv = lin::ref<3, 1>(q, 0, 0);
+  res = v + lin::cross(
+      (static_cast<T>(2.0) * qv).eval(),
```
nhz2, line 41: Could you explain how lin::cross works? Sometimes you are using `.eval()` on the returned value, but sometimes not. Whats the difference?

.travis.yml
kylekrol: Honestly I don't really understand how it works either haha. I just looked at some examples online. It seems like you can run multiple 'instances' independently of one another making it easy to write separate CI pipelines for different languages.

include/inl/gnc_utilities.inl
kylekrol: Lin already does that type checking. Currently it only enable the `lin::internal::traits<...>` struct if `T` is a float or double (only supported types currently).

**EDIT**: For reference here's a link https://github.com/kkrol27/lin/blob/ada842a53b1dfbb6a50676464e36a1041f8e8576/include/core/tensor/traits.hpp#L24
The struct's value is checked by `is_traits_valid` as well which is asserted in every `lin::internal::Stream<>` class (which is literally every type in the library).

include/inl/gnc_utilities.inl
kylekrol: The `*.eval()` calls force the creation of a temporary object and prevents lazy evaluation. For example, writing `c = a + b` actually equates, in terms of types to `lin::Vector<> = lin::internal::Add<lin::Vector<>, lin::Vector<>>` and then a single for loop performs the addition and assignment all in one big swoop. In contrast writing `c = (a + b).eval()` gives us a temporary `lin::Vector<>` on the RHS and then a second for loop to perform assignment from the temporary to `c`.

The big picture here is that most functions buried in the library actually take in a "stream" which can be a lazily evaluated operator proxy, a read only reference, read/write reference, or actual vector/matrix object. As the programmer you need to be aware when temporaries should be or shouldn't be created. In the above case, forcing the evaluation of `static_cast<T>(2.0) * qv` saved three floating point ops because the cross product using each component's value twice.

Hopefully that makes sense. I apologize for the documentation being sparse.

test/utilities_test.cpp
kylekrol: I think we are pretty well guarded against that for the time being. Especially when it comes to the functions added in this PR - there isn't any scalar division stuff that could result in `NaN`.

include/inl/gnc_utilities.inl
```diff
@@ -0,0 +1,138 @@
...
+  lin::Vector<T, 3> const  n2 = lin::cross(n1, N_mag) / lin::norm(lin::cross(n1, N_mag));
+  lin::Vector<T, 3> const  n3 = lin::cross(n1, n2);
+
+  // Calculate right handed bases for the body frame
+  lin::Vector<T, 3> const &b1 = B_sun;
+  lin::Vector<T, 3> const  b2 = lin::cross(b1, B_mag) / lin::norm(lin::cross(b1, B_mag));
+  lin::Vector<T, 3> const  b3 = lin::cross(b1, b2);
+
+  // Calculate the DCM between the two frames and extract the quaternion
+  auto const Q = b1 * lin::transpose(n1) + b2 * lin::transpose(n2) + b3 * lin::transpose(n3);
```
kylekrol, line 133: Just as another example, the auto keyword here actually represents a type like `Add<Multiply<Vector<>, Transpose<Vector<>>>, Add< ... >>` and the computation to calculate the result is done by `Q.eval()` on the next line.

PR #43, opened 2019-10-21
perturbing force due to atm drag

srutiv:
Relates to #8


MATLAB/dynamics_update.m
```diff
@@ -98,10 +98,12 @@
         g_eci=rotateframe(quat_eci_ecef,g_ecef')';
         %TODO add thruster firings forces and torques
         %TODO add drag and solar pressure forces
-        dydt(4:6)= g_eci;
+        F_envdrag = env_atmospheric_drag(t,y(1:3),y(4:6)); %drag force; input mission time in sec, position ECI m, and velocity ECI m/s
+        dydt(4:6)= g_eci + F_envdrag/(const.MASS+fuel_mass(t));
```
kylekrol: I would say we should probably calculate `F_envdrag` just once outside of the function passed to our integrator. The drag force will most likely be constant over the course of our integration intervals and that would help us with performance.

This was done with the magnetic field at the location of the spacecraft on line 76.

@nhz2 @stewartaslan What do you think about this?

kylekrol (commented):
Looks great other than that one concern - nice job!

MATLAB/environmental_models/env_atmospheric_drag.m
```diff
@@ -0,0 +1,89 @@
...
+UTC_leap = utl_time2datetime(time,const.INITGPS_WN);  %converts mission time in secs to datetime in UTCleapseconds
+%UTC_leap = utl_time2datetime(time,3); test
+offset = 2*86400; %offset of UTCLeapSeconds to UTC time
+del = datetime(0,0,0,0,0,offset);
+del.TimeZone = 'UTCLeapSeconds';
+UTC = UTC_leap - offset;
+
+% r: satellite position in ECI
+%convert ECI coordinates to latitude, longitude, altitude (LLA) geodetic coordinates based on the Universal Coordinated Time (UTC) you specify
+lla = eci2lla(r, [UTC.Year UTC.Month UTC.Day UTC.Hour UTC.Minute UTC.Second]);
```
nhz2: you have to transpose r so its a row vector, like `r'` to use `eci2lla`.

MATLAB/environmental_models/env_atmospheric_drag.m
```diff
@@ -0,0 +1,89 @@
...
+h = abs(lla(3)); %altitude of Position (m); geocentric altitude = geodetic altitude
+
+rho = get_rho(h);
+
+Cd = 0.8; %drag coefficient of angled cube
+A = 0.1*sqrt(2)*0.3; %largest planar area of satellite in m^2
+v_rel = v - cross(const.earth_rate_ecef,r); %velocity relative to the rotating atmosphere
+%v_rel = v - cross([0;0;7.2921158553E-5],r); test
+
+F_envdrag = -0.5*rho*Cd*A*(v_rel*v_rel'); %drag calculated in eci frame
```
nhz2: `v_rel` is a column vector as well, so i think this should be `v_rel'*v_rel`.

MATLAB/environmental_models/env_atmospheric_drag.m
```diff
@@ -0,0 +1,89 @@
+function F_envdrag = env_atmospheric_drag(time,r,v)
+%input: mission time in sec
+% r position in ECI frame in m
```
nhz2, line 3: Input position in the ECEF frame, instead of ECI frame. Then you can use, `ecef2lla` instead of `eci2lla` which is much faster and you won't need to deal with all of the UTC leap second, datatime stuff.

nhz2: Actually, maybe input both, because you need position in ECI for `v_rel = v - cross(const.earth_rate_ecef,r);` to calculate velocity relative to the rotating atmosphere?

MATLAB/environmental_models/env_atmospheric_drag.m
jer296: It looks like F_envdrag is a scalar. Should it be a vector?
F_envdrag = -0.5*rho*Cd*A*(v_rel*v_rel')*(v/norm(v));

jer296: F_envdrag = -0.5rhoCdA(v_relv_rel')(v_rel/norm(v_rel));
Actually use v_rel instead of v for the vector direction

MATLAB/environmental_models/env_atmospheric_drag.m
srutiv: updated

MATLAB/environmental_models/env_atmospheric_drag.m
srutiv: updated

MATLAB/environmental_models/env_atmospheric_drag.m
srutiv: updated

PR #49, opened 2019-10-22
Made the get_truth function.

nhz2:
Call `get_truth("help")` in matlab to see the current supported values. Currently the `'lvlh'`, and `'vbn'` frames don't have angular rates.


MATLAB/get_truth.m
```diff
@@ -0,0 +1,236 @@
+function value = get_truth(name,dynamics)
```
tanishqaggarwal, line 1: This requires a corresponding change in the USB console to grab truth from this function rather than from the truth variable, right?

nhz2: Yeah, you can still get the core data directly from dynamics, but it might be nice to let the user request additional data in any frame they want using `get_truth`.

stewartaslan (approved):
Looks great! Could you add a few comments in the get_truth.m function itself? It's very dense.

Issue #48, opened 2019-10-21
RadioSession implementation


tanishqaggarwal:

tanishqaggarwal:
Implementing in #51

PR #51, opened 2019-10-25
Dev/gsw

fatimayousuf:
Updated the radio session class. Addresses #48.


python/usb_console/radio_session.py
```diff
@@ -36,8 +41,16 @@ def __init__(self, device_name, datastore, logger):
...
 
-        # Simulation
-        self.overriden_variables = set()
+        #email        
+        with open(os.path.join(os.path.dirname(os.path.abspath(__file__)), 'configs/confidential.json')) as config_file:
+            data = json.load(config_file)
+        self.username=data["PANUsername"]
+        self.password=data["PANPassword"]
+
+        self.statefields = {
```
tanishqaggarwal: Can you remove these testing values?

python/usb_console/radio_session.py
```diff
@@ -59,27 +72,78 @@ def check_for_downlink(self):
...
+                                continue
+                            
+                            #check if there is an attachment
+                            if part.get_filename() is not None:
+                                #get json data from email attachment
+                                attachmentContents=part.get_payload(decode=True).decode('utf8')
+                                data=json.loads(attachmentContents)
+                                
+                                #iterate through each statefield key of the dictionary
+                                logline=""
```
tanishqaggarwal: Could you instead log the state field value in the datastore? The logger is a good place to put debugging data (e.g. "Connected to Email!", "Disconnected from email.") but the datastore should be the destination for any telemetry collected by a Session.

When you put a value into the datastore, the value should be dict with two keys: "time", which is the datetime of the received value, and "val", which is the actual value.

python/usb_console/radio_session.py
```diff
@@ -59,27 +72,78 @@ def check_for_downlink(self):
...
+                                continue
+
+                            if part.get('Content-Disposition') is None:
+                                continue
+                            
+                            #check if there is an attachment
+                            if part.get_filename() is not None:
+                                #get json data from email attachment
+                                attachmentContents=part.get_payload(decode=True).decode('utf8')
+                                data=json.loads(attachmentContents)
```
tanishqaggarwal: It would be nice if the processing of the attachment data into a dictionary happened in a separate function. Right now, for testing purposes, we're assuming the data in the attachment is JSON, but in the future it'll be a really complicated encoding scheme. Maybe you could create a new function called process_downlink_packet(), and inside it call json.loads on the attachment contents? This also makes the downlink-processing part of the logic more easily unit-testable in the future.

python/usb_console/radio_session.py
```diff
@@ -59,27 +72,78 @@ def check_for_downlink(self):
...
 
         while self.running_logger:
-            # TODO implement
-            pass
+            #.search() searches from mail. Data gives id's of all emails.
+            type, data = mail.search(None, '(FROM "sbdservice@sbd.iridium.com")', '(UNSEEN)')
+            mail_ids = data[0]
+            id_list = mail_ids.split()
+
+            for num in id_list:
```
tanishqaggarwal: Can you also filter messages by IMEI number? The IMEI number is the subject line of the email.

The reason for this is that the PAN email address will receive data from both the follower and leader spacecraft, and a radio session should be able to distinguish between them.

Related: could you tell me how this code marks an email as unread?

python/usb_console/data_consumers.py
```diff
@@ -94,3 +95,6 @@ def save(self):
         filepath = os.path.join(self.data_dir, filename)
         with open(filepath, 'w') as logfile:
             logfile.write(self.log)
+    
+    def put(self, logline):
+        self.queue.put("("+str(datetime.now())+")"+"\n"+logline+"\n")
```
tanishqaggarwal: Because of this change the unit test for this class now fails (you can see this in the Travis CI build). I forgot that changing this would break it. Could you fix it?

python/usb_console/run_simulation.py
```diff
@@ -118,7 +119,7 @@ def set_up_radios(self):
                 radio_session = RadioSession(radio_connected_device, radio_datastore, radio_logger)
 
                 if radio_session.connect(radio['imei']):
```
tanishqaggarwal: It would also be good if the radio IMEI number came from your confidential.json file, rather than the config files that are currently publicly available. Can you modify the run_simulation logic for this?

python/usb_console/radio_session.py
```diff
@@ -36,8 +41,16 @@ def __init__(self, device_name, datastore, logger):
         self.datastore = datastore
         self.logger = logger
 
-        # Simulation
-        self.overriden_variables = set()
+        #email        
+        with open(os.path.join(os.path.dirname(os.path.abspath(__file__)), 'configs/confidential.json')) as config_file:
```
tanishqaggarwal: Can you rename confidential.json to something like "radio_keys.json"? Makes the file more descriptive.

python/usb_console/radio_session.py
```diff
@@ -36,8 +41,16 @@ def __init__(self, device_name, datastore, logger):
...
+            data = json.load(config_file)
+        self.username=data["PANUsername"]
```
tanishqaggarwal: Same deal here with description: I think it would be better if this is called the "email_username" and "email_password".

python/usb_console/radio_session.py
```diff
@@ -59,27 +72,78 @@ def check_for_downlink(self):
...
+            mail_ids = data[0]
+            id_list = mail_ids.split()
+
+            for num in id_list:
+                #.fetch() fetches the mail for given id where 'RFC822' is an Internet 
+                # Message Access Protocol.
+                typ, data = mail.fetch(num,'(RFC822)')
+                
+                #go through each component of data and print sender and subject line
+                for response_part in data:
```
tanishqaggarwal: It might be better to split off the actual email processing into a different function in the class--there's a lot of nested for-loops here.

python/usb_console/radio_session.py
```diff
@@ -59,27 +72,78 @@ def check_for_downlink(self):
...
     def write_multiple_states(self, fields, vals, timeout=None):
         '''
         Uplink multiple state variables. Return success of write.
         '''
         assert len(fields) == len(vals)
 
-        # TODO send uplink via email.
-        raise NotImplementedError
+        #update statefield dictionary
+        for i in range(len(fields)):
```
tanishqaggarwal: This is actually not what we want to do. We're not going to be sending all of the state fields up to the satellite; we'll only send the ones specified in the arguments of this function. I would remove this for loop and instead construct a new dictionary based only on `fields` and `vals`.

python/usb_console/radio_session.py
```diff
@@ -59,27 +72,78 @@ def check_for_downlink(self):
...
     def read_state(self, field, timeout=None):
         '''
         Read state.
         
         Read the latest value of the state field received via downlink with the name "field".
         '''
-        raise NotImplementedError
+        return self.statefields[field]
 
     def write_multiple_states(self, fields, vals, timeout=None):
```
tanishqaggarwal: In the future, we want to revisit this function to add the following features:
- Wait for a confirmation email from Iridium that the Iridium satellite network received the packet. The email returns an MTMSN number, which stands for "Mobile-Terminated Message Number" number. This number should be saved somewhere as a member variable of this class because:
- If you check out some of the emails from Iridium containing downlinks, you'll see that they contain a line starting with MTMSN. This number corresponds to the latest uplink processed by Quake. If this number isn't equal to the MTMSN of the most recent uplink sent to the Quake, then this class should forbid its user from sending additional uplinks. The goal of this is to prevent more than one command-carrying uplink packet to be processed by the satellite at a time.

For the last bullet point, we'll need to add some email body-reading functionality to `check_downlink_thread` as well.

Can you start thinking about how to include these features into the logic?

tanishqaggarwal: One more point; we want to store the MOMSN number that's received in the body of a downlink email as well, so that we have a record of it.

python/usb_console/radio_session.py
```diff
@@ -153,6 +157,9 @@ def read_state(self, field, timeout=None):
         Read state.
         
         Read the latest value of the state field received via downlink with the name "field".
+
+        I can get the latest MTMSN value from the confirmation email through the check downlink function (self.mtmsn).
+        How do I get MTMSN number of the latest uplink?
```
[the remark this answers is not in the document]
tanishqaggarwal: Might be a good idea to rename `check_for_downlink` to `check_email` in this case, since the thread would have to do the job of processing both uplink as well as downlink emails. Try to keep everything as logically named as possible when making this change--shoehorning in features reduces readability.

python/usb_console/radio_session.py
```diff
@@ -54,32 +77,110 @@ def connect(self, imei):
         print(f"Opened connection to {self.device_name} radio.")
         return True
 
+    def process_downlink_packet(self, data):
```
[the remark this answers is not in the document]
tanishqaggarwal: Can you also add some more documentation as to why this function exists? Something like "This is just deserializing JSON for now, but in the future will be replaced with the actual deserialization scheme for downlink packets."

python/tests/test_data_consumers.py
```diff
@@ -25,9 +25,10 @@ def test_logger():
...
     logger.start()
     logger.put("Hello world")
+    currentTime=str(datetime.datetime.now())
     time.sleep(1.5) # Allow for some time for the data to be processed by the queue processor
     logger.stop()
 
     with open(tempdir + "/test_device-log.txt", "r") as logfile:
         log = logfile.read()
-        assert log == "Hello world\n"
+        assert log.find("Hello world\n") != -1
```
tanishqaggarwal: A cleaner way to write this is `assert "Hello world" in log` but this is fine too

Issue #52, opened 2019-10-27
Add utl_vect_rot2quat and unit tests (including NaN test cases)


nhz2:
function qout = utl_vect_rot2quat(r1,r2)
%UTL_VECT_ROT2QUAT Returns a unit quaternion that will rotate r2 to r1 with minimal angle.
%   If r1 and r2 are very close to anti parallel, [0;0;1;0] is default.
%     Quaternions have the forth component as scalar.
%     r1 and r2 must unit vectors or have NaN or +- inf
%   the output becomes nonfinite if any input is nonfinite

kylekrol:
Actually have this done on a branch right now - however they don't check for NaN stuff. Do we really want the responsibility of checking whether or not something is NaN to a lower level utility function?

nhz2:
Yeah, low level utility functions should usually not need to check for NaN, propagation is automatic. Just make sure the code still runs if it has NaN input, and that the quaternion output becomes nonfinite if any input is nonfinite. I'll add that to the Matlab function documentation.

tanishqaggarwal:
@kkrol27: closing because the functions are already implemented, the NaN tests weren't done but we're using them at a FlightSoftware level anyway, so we have enough protection.

PR #53, opened 2019-10-27
Updates to GNC utility functions and the CMake build system

kylekrol:
Relates to #52.

Thought I'd make this intermediate PR before starting the CXX estimation function as it make take a little longer and I thought it best to get these other updates in first.

Major things in the PR
 * `./test/*` renamed to `./tests/*` for annoying CMake target reasons
 * `CI_BUILD` macro flag is now supported so tests/code can be written only for use locally (could be helpful for tests using random input that have a slim chance of failing)
 * `vec_rot_to_quat` utility function was added along with unit tests (`NaN` support in CXX is yet to come - I gotta do a little more reading into compiler flags and what not)
 * Unit tests for all `dcm_to_quat` cases were added.

Lastly, yes the `gnc_state_estimation.*pp` files are empty. However it needs to be there in the beginning for the build system and will get filled in soon in a following PR.


tanishqaggarwal:
If writing tests with random input, I thought the better way to handle CI failures would be to feed the randomness with a set of known-good fixed seeds? This would ensure the tests always run in CI but never fail if they don't fail even once locally.

Everything else looks good. Looking forward to seeing that gnc_state_estimation.hpp and then I can start writing the necessaries on the flight software side.

kylekrol:
Oooh yeah the seeding thing is a good idea - thanks!

nhz2 (approved):

Issue #54, opened 2019-10-27
actuator_commands = actuators_off_command()


nhz2:
Constructs an `actuator_commands` that commands all of the actuators off. 

`actuator_commands` is a struct with elements:
firing_start_times(4x1 matrix): commanded times since initial GPS week to start firing (s)
firing_on_times(4x1 matrix): commanded thruster on times (s)
wheel_torque, commanded x,y,z wheel torque, (signed ramp)x(rotor inertia) (Nm)
wheel_enable, commanded x,y,z wheel enables, whether each wheel should be on, if false, the wheel rate is commanded to zero.
magrod_moment, commanded x,y,z magnetorquer moments (Am^2)

PR #55, opened 2019-10-27
Create actuators_off_command.m

nhz2:
This function creates an `actuator_commands` struct  that turns off all of the actuators. It can also be used as a default constructor for `actuator_commands` on the python side, to get a list of all the needed keys and values. #54


kylekrol (approved):
Looks good to me!

Issue #47, opened 2019-10-21
Add J4/J8 Orbit Propagator


kylekrol:
Depends on #65.

We need to code a J4/J8 orbit propagator in C++ to fly on the spacecraft. This will required an integrator implementation like `ode45` to be included in PSim as well.

tanishqaggarwal:
Along with this issue (since it directly relates to FSW): we need to figure out a painless way to share code between psim and FlightSoftware.

I think importing psim as a library into FlightSoftware would be ideal, but this requires being cognizant about the fact that platformio only finds library files if they're in the root-level of src/ and include/. This might be tricky (or not) given that `lin` has already been placed in a place that PlatformIO might not be able to find it--have you been able to test this or know about the inclusion structure?

The other option would be to add psim as a submodule to FlightSoftware. I'm not a fan of this because it makes the directory structure even more convoluted and difficult to import into PlatformIO. Unless I'm imagining a difficulty where it doesn't actually exist.

I think a first step would be to create stub files for things we anticipate absolutely needing in FSW, and experimenting with FlightSoftware and platformio includes until everything flows together. What do you think?

tanishqaggarwal:
Nevermind. I forgot that we're already tracking this integration on #41 .

tanishqaggarwal:
Duplicate of #78

PR #62, opened 2019-11-04
Add Fixed Step Size Integrators

kylekrol:
Relates to #47 (and many other things to come).

This includes the `ode1`, `ode2`, `ode3`, and `ode4` integrators. Tests were also added for compare the CXX function outputs with MATLAB's integrators for the simple harmonic oscillator problem. The integrator coefficients used were identical to MATLAB's so the CXX and MATLAB output should be pretty much identical.

Here is the MATLAB script used to generate the expected outputs:

    format long
    t = [0.0, 0.1, 0.2, 0.3];
    y = [1.0; 0.0];

    disp 'ode1'
    utl_ode1(@f, t, y)
    disp 'ode2'
    utl_ode2(@f, t, y)
    disp 'ode3'
    utl_ode3(@f, t, y)
    disp 'ode4'
    utl_ode4(@f, t, y)

    function dy = f(~, y)
        [r, c] = size(y);
        dy = zeros(r, c);
        dy(1) = y(2);
        dy(2) = -y(1);
    end

I also removed windows specific support from the CMake file (for Visual Studio) as I'm not familiar with the compiler flag options and I don't think anyone is working on the CXX in this repo with visual studio. If this changes, adding support shouldn't be too difficult.

### DIFF
--- a/tests/ode_test.cpp
+++ b/tests/ode_test.cpp
@@ -1,0 +1,81 @@
+//
+// test/utilities_test.cpp
+// PSim
+//
+// Contributors:
+//   Kyle Krol  kpk63@cornell.edu
+//
+// Pathfinder for Autonomous Navigation
+// Space Systems Design Studio
+// Cornell Univeristy
+//
+
+#include <gnc_ode.hpp>
+
+#include <gtest/gtest.h>
+
+#define PAN_GNC_ODE_TEST_VARS(n) \
+    unsigned int const nt = 4; \
+    unsigned int const ne = 2; \
+    double const t[nt] = { 0.0, 0.1, 0.2, 0.3 }; \
+    double bf[n * ne]; \
+    double _y[nt * ne]; \
+    double *y[nt]; \
+    for (unsigned int i = 0; i < nt; i++) \
+      y[i] = _y + ne * i; \
+    y[0][0] = 1.0; \
+    y[0][1] = 0.0;
+
+static void fsho(double t, double const *y, double *dy) {
+  dy[0] = y[1];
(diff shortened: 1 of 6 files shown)


tests/ode_test.cpp
```diff
@@ -0,0 +1,81 @@
...
+static void fsho(double t, double const *y, double *dy) {
+  dy[0] = y[1];
+  dy[1] = -y[0];
+}
+
+TEST(OdeTest, Ode1ShoTest) {
+  PAN_GNC_ODE_TEST_VARS(1)
+  gnc::ode1(t, nt, y, ne, bf, fsho);
+  // Check against MATLAB output
+  ASSERT_FLOAT_EQ(y[1][0],  1.000000000000000);
```
tanishqaggarwal, line 38: I'm amazed this manages to pass, considering rounding errors. Would it be a good idea to add a tolerance for the equality?

tanishqaggarwal (approved):
The code is very difficult to read, but in a way that's fundamentally unavoidable with numerical methods. Tests look OK and so I'm not too concerned, since we'll be doing plenty of numerical verification atop these methods.

tests/ode_test.cpp
kylekrol: I mirrored the MATLAB implementations and wrote the tests in doubles so we could be this picky with the tolerances. Gtest has some type of very tight tolerances built in for the `ASSERT_FLOAT_EQ` macro; although, I don't remember of the top of my head what it is. If we start seeing issues down the line we can revisit.

PR #63, opened 2019-11-04
Add Environment Functions and Unit Tests

kylekrol:
Relates to #47 (and many other things).

@nhz2 @stewartaslan @tanishqaggarwal Opening up a quick draft here for awareness. The build will be broken for this branch due to a `-Wextra` unused parameter error. I had to disable this for integrator stuff as well so I'll wait for that other PR to be merged.

I've started to code more of the environmental models into C++ as we'll need them to write the attitude/state estimation function in C++. From a brief run through of the current MATLAB, it seems like the best way forward is to have a `gnc_constants.hpp` file that constants values like earth's precession rate, perihelion time, orbit eccentricity, etc etc as a lot of the environmental models and GNC functions will run off of them. FSW can copy them over to somewhere else if it needs them and I think that should be fine.

@nhz2 @tanishqaggarwal does this seem reasonable to both of you?

@nhz2 I have a couple questions I will ask you offline about some of the math specifics. I got myself slightly turned around while writing this and want to nail down some notational conventions.

**EDIT**:

I've updated this PR to include the sun vector environmental model as well as the magnetic field model. All models are working with unit tests in cxx and the sim runs well (I moved the geomap header into `src` and hid it in an `inl` folder as it's more of an implementation detail).


include/gnc_constants.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+namespace gnc {
+namespace constant {
+
+/** Initial GPS week number that serves as the 'PAN epoch'. Some of the
+ *  constants below depend on this number. */
+constexpr static unsigned int init_gps_week_number = 2045;
+
+/** Earth's intertial rotation rate about the z axis in the ECEF frame.
+ *  NOTE: Pulled from PSim. */
```
tanishqaggarwal: Is it worth adding more specific references here? Someone like me who doesn't know psim wouldn't know where to look

src/gnc_environment.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+#include <cmath>
+
+namespace gnc {
+namespace env {
+
+void earth_attitude(double t, lin::Vector4d &q) {
+  using namespace lin;
+  // Determine a transformation from ecef0 to ecef0p
+  Vector4d q_ecef0p_ecef0({
```
tanishqaggarwal: What's the difference between ECEF0P and ECEF0?

include/gnc_constants.hpp
```diff
@@ -0,0 +1,52 @@
...
+#define PAN_PSIM_INCLUDE_GNC_CONSTANTS_HPP_
+
+#include "lin.hpp"
+
+namespace gnc {
+namespace constant {
+
+/** Initial GPS week number that serves as the 'PAN epoch'. Some of the
+ *  constants below depend on this number. */
+constexpr static unsigned int init_gps_week_number = 2045;
```
tanishqaggarwal: Can you add GPS time of week and nanoseconds here as well?

src/gnc_environment.cpp
```diff
@@ -0,0 +1,61 @@
...
+// Pathfinder for Autonomous Navigation
+// Space Systems Design Studio
+// Cornell Univeristy
+//
+
+#include <gnc_constants.hpp>
+#include <gnc_environment.hpp>
+#include <gnc_utilities.hpp>
+
+#include <cmath>
```
tanishqaggarwal, line 17: Why is this include necessary?

include/gnc_environment.hpp
```diff
@@ -0,0 +1,44 @@
...
+
+/** @fn earth_attitude
+ *  @param[in]  t Time in seconds since the PAN epoch.
+ *  @param[out] q Quaternion to transform vectors from ECI to ECEF. */
+void earth_attitude(double t, lin::Vector4d &q);
+
+/** @fn earth_attitude
+ *  @param[in]  t Time in seconds since the PAN epoch.
+ *  @param[out] q Quaternion to transform vectors from ECI to ECEF. */
+void earth_attitude(double t, lin::Vector4f &q);
```
tanishqaggarwal, line 29: What's the utility of these float-based functions? Wouldn't it be better to do all of the arithmetic with doubles, and then convert to floats only if necessary for downlinking?

include/gnc_constants.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+/** Earth's intertial rotation rate about the z axis in the ECEF frame.
+ *  NOTE: Pulled from PSim. */
+constexpr static double earth_rate_ecef_z = 1.0e-04 * 0.729211585530000;
+
+/** Approximate precssion rate of Earth's Axis in radians per second.
+ *  NOTE: Pulled from PSim and needs to be updated with a new GPS week. */
```
tanishqaggarwal: Can you create an issue tracking these "required updates" so that we don't lose them

tanishqaggarwal (changes requested):
Mostly questions, some actionable items

src/gnc_environment.cpp
kylekrol: `sin` and `cos`.

include/gnc_constants.hpp
kylekrol: Yep I was going to flush all of this out and potentially invest the time in writing a script to autocoder the required constants and unit tests. Wanted to discuss this stuff in more detail, however, and we'll probably get to that Thursday would be my guess.

include/gnc_environment.hpp
kylekrol: The only thing would just be wasting instructions if we use doubles all the time (although that may not be a huge deal). For the moment, I've been trying to to specialize because we aren't exactly sure what the controllers are going to look like.

For example, I know we want to do attitude stuff in floats because we don't need the accuracy. However, the time here clearly needs to be passed as a double for precision so we can be multiplying that double onto a float before normalizing the quaternion. However, once the quaternion is normalized our precision issues disappear. So in that case, the float function has a utility.

tanishqaggarwal: Yeah I suppose it's not a huge deal to have them, either, but eventually we want to be careful about having too many functions to deal with

src/gnc_environment.cpp
kylekrol: It's sort of a long story that @nhz2 can explain better. But essentially you're treating the Earth's precession and rotation independently from each other and therefore get these intermediate frames.

So you go ECI -> ECEF0 (ECEF at time 0) -> ECEF0p (no precession just z axis rotation) -> ECEF (precession and z axis rotation) or something along those lines.

include/gnc_constants.hpp
kylekrol: Yeah no for sure that was the plan. This was just a very early mockup to make sure I wasn't putting time into something that we thought should be done another way.

include/gnc_constants.hpp
kylekrol: Yep for sure we will be adding more constants. How we handle time specifically is something I think I'm lagging on. It's on my list of potential topics for Thursday.

tanishqaggarwal: Since GPS epoch is a specific constant that FSW requires I think it'd be good if we can get it in on this PR

include/gnc_constants.hpp
```diff
@@ -0,0 +1,81 @@
...
+
+/** Good old pi. */
+constexpr static double pi = 3.141592653589793;
+
+/** Two pi. */
+constexpr static double two_pi = 2.0 * pi;
+
+/** Initial GPS week number that serves as the 'PAN epoch'. The 'PAN epoch'
+ *  aligns with a GPS week so no additional offset is required. */
+constexpr static unsigned short init_gps_week_number = 2045;
```
kylekrol, line 42: @tanishqaggarwal It wouldn't let me continue the other conversation for some reason. To answer your earlier question however about adding in a nanosecond count for the 'PAN epoch', I'm pretty sure we're setting an epoch at the very start of a GPS week so the initial nanosecond count is zero.

@nhz2 could you confirm.

tanishqaggarwal: I agree it should be zero, but for the sake of having a named constant, I think it would be a good idea to have the TOW and nanoseconds explicitly defined?

include/gnc_environment.hpp
kylekrol: Internally we will have a lot of functions most likely. But the interface with the flight computer will hopefully be very narrow and we shouldn't have functions overloaded just for `float` vs `double`.

src/gnc_environment.cpp
```diff
@@ -0,0 +1,102 @@
...
+void earth_attitude(double t, lin::Vector4f &q) {
+  lin::Vector4d _q;
+  earth_attitude(t, _q);
+  for (lin::size_t i = 0; i < q.size(); i++)
+      q(i) = static_cast<float>(_q(i));
+}
+
+void earth_angular_rate(double t, lin::Vector3d &w) {
+  w = { 0.0, 0.0, constant::earth_rate_ecef_z };
+}
```
kylekrol, line 56: @nhz2 I was curious about this bit of code from MATLAB. Doesn't this neglect the `w` contribution from the Earth's precession? Are we just calling it negligible because it's so small relative to the rotation about the z axis?

include/gnc_constants.hpp
kylekrol: Alright yeah that's fair just in case we do something goofy down the road. Will add it in!
**EDIT**: That little bit should be good to go then I think.

tanishqaggarwal (dismissed):

include/gnc_constants.hpp
```diff
@@ -0,0 +1,87 @@
...
+#ifndef PAN_PSIM_INCLUDE_GNC_CONSTANTS_HPP_
+#define PAN_PSIM_INCLUDE_GNC_CONSTANTS_HPP_
+
+#include "lin.hpp"
+
+namespace gnc {
+namespace constant {
+
+/** Good old pi. */
+constexpr static double pi = 3.141592653589793;
```
nhz2: I would add the `L` suffix to all the double constants, just to make sure they don't get compressed for teensy. [https://forum.pjrc.com/threads/47790-double-literals-need-to-be-marked-as-long-double](https://forum.pjrc.com/threads/47790-double-literals-need-to-be-marked-as-long-double)

include/gnc_environment.hpp
```diff
@@ -0,0 +1,68 @@
...
+
+/** @fn earth_angular_rate
+ *  @param[in]  t Time in seconds since the PAN epoch.
+ *  @param[out] w Angular rate to transform vectors from ECI to ECEF. */
+void earth_angular_rate(double t, lin::Vector3f &w);
+
+/** @fn sun_vector
+ *  @param[in]  t Time in seconds since the PAN epoch.
+ *  @param[out] s Normalized vector from Earth the Sun in the ECI frame. */
+void sun_vector(double t, lin::Vector3d &s);
```
nhz2: I'm pretty sure we don't need a double version of this.

include/gnc_environment.hpp
```diff
@@ -0,0 +1,68 @@
...
+ *  @param[out] s Normalized vector from Earth the Sun in the ECI frame. */
+void sun_vector(double t, lin::Vector3f &s);
+
+/** @fn magnetic_field
+ *  @param[in]  t Time in seconds since the PAN epoch.
+ *  @param[in]  r Position in the ECEF frame with units of meters.
+ *  @param[out] b Magnetic field in the ECEF frame with units of Tesla.
+ *  NOTE: The double overload of this function provides no extra accuracy over
+ *        the float version. */
+void magnetic_field(double t, lin::Vector3d const &r, lin::Vector3d &b);
```
nhz2: We definitely don't need a double version of magnetic_field

tests/environment_test.cpp
```diff
@@ -0,0 +1,127 @@
...
+    -0.013041330575474,
+  }));
+}
+
+TEST(EnvironmentTest, MagneticFieldTest) {
+  lin::Vector3f b, r;
+  // Calculate and check for t=0.0 (comparing against MATLAB)
+  r = { 5.430929361985142e6f, 2.949986114465339e6f, 2.949986114465338e6f };
+  gnc::env::magnetic_field(0.0, r, b);
+  ASSERT_VEC_NEAR(1e-5f, b, lin::Vector3f({
```
nhz2: The margin for this test should be around 10E-9 Teslas

include/gnc_constants.hpp
kylekrol: Oooh okay thanks! Will do!

include/gnc_environment.hpp
kylekrol: Getting rid of it

include/gnc_environment.hpp
kylekrol: I'm going to leave it for the time being I think simply because we need to do the calculation in doubles up until the quaternion is normalized. Otherwise we'd risk losing a lot of precision once `t` gets large.

tanishqaggarwal (dismissed):

include/gnc_environment.hpp
kylekrol: Got rid of the double version after realizing we'd need to be in orbit for 100's of years for accuracy to become a problem. By that point the approximations made by this function would already fall apart anyway.

nhz2 (approved):

PR #73, opened 2019-11-06
Transition the Unit Test and Build System to PlatformIO

kylekrol:
Creating to run CI for the time being.


test/environment_test/test.cpp
```diff
@@ -24,104 +24,108 @@ static_assert(gnc::constant::init_gps_time_of_week == 0,
...
-
-TEST(EnvironmentTest, EarthAttitudeTest) {
+void test_env_earth_attitude() {
   lin::Vector3d w;
   lin::Vector4d q;
   // Calculate and check for t=0.0 (comparing against MATLAB)
   gnc::env::earth_attitude(0.0, q);
   gnc::env::earth_angular_rate(0.0, w);
-  ASSERT_VEC_NEAR(1e-7, q, lin::Vector4d({
+  TEST_ASSERT_DOUBLE_VEC_NEAR(1e-7, q, lin::Vector4d({
```
tanishqaggarwal, line 33: Now that I think about it--I'm wondering if there may be numerical issues that cause lin to fail on Teensy? Is that possible?

test/test.hpp
```diff
@@ -0,0 +1,49 @@
...
+// Contributors:
+//   Kyle Krol  kpk63@cornell.edu
+//
+// Pathfinder for Autonomous Navigation
+// Space Systems Design Studio
+// Cornell Univeristy
+//
+
+#ifdef PAN_PSIM_TEST_TEST_HPP_
+#error test.hpp should not be included twice
```
tanishqaggarwal, line 14: Why are you erroring here?

.travis.yml
```diff
@@ -1,19 +1,22 @@
...
+      python: 2.7
+      sudo: false
+      cache:
+        directories:
+          - "~/.platformio"
+      install:
+        - pip install -U platformio
+        - platformio update
+      script:
+        - platformio test --verbose --environment ci_native
```
tanishqaggarwal: Can you add `platformio run` as part of the build script? This ensures that all code compiles, at least, before test cases are run

tanishqaggarwal (changes requested):

.travis.yml
kylekrol: There is no `main` or `setup` `loop` functions so the build will just fail. The test functions builds all code, however, and will spit out the build error if one occurs.

test/environment_test/test.cpp
kylekrol: The tests failing don't even use lin - currently only `ode23` is failing. It has to come down to a compiler difference between clang and GNU. I've had my friend build on different laptops and it only fails with GNU compilers - really beats me why.

test/test.hpp
kylekrol: This code is supposed to be an extension to a single cpp file. If you're including it twice you're doing something wrong lol. It is a little weird for sure. I just hacked this together to get stuff to build and up and running with CI.

kylekrol: Yeah idk - I've been looking at this for a while now. I feel weird giving it a standard header guard when the code in it behaves far different from a normal header file (`main` is even defined inside it!)

kylekrol:
While this didn't fix the precision issue unfortunately, I think these changes are important to merge with master ASAP. I will open up a separate ticket to track the GCC compiler problem.

tanishqaggarwal (approved):

test/test.hpp
tanishqaggarwal: Oh, now I get it. This is actually a really cool setup; it minimizes overhead.

PR #76, opened 2019-11-08
Fixed Ode23 Precision Issue with GCC

### LINKED ISSUE #74: Debug Why Ode23 Testcase Fails when Built with GCC Compilers
Relates to #67 and #71.

It seems that the `ode23` test in the current version of PSim will fail on all GCC compilers yet pass when built with clang. This failure has been seen on Ubuntu, Mint, the CI servers, and PlatformIOs GCC ARM compiler. The test, however, passes on Mac with Clang - Clang has not been tried on other platforms yet.

It's important to note this is by no means small difference. On my machine locally, the error between expected and calculate in `~1e-6` and on the failing platforms it's `~0.005`.

Going forward, I first want to set up a Docker container to give everyone (relatively) easy acmes to a Linux development environment. From there we can start to debug the issue much more efficiently.

kylekrol:
Closes #74.

This ended up being a super simple fix (lucky for us). I'm not 100% sure I totally understand what's going on but here seems to be the short version.

`#include <cmath>` also pulls in an `#include <cmath.h>`. This has the implication that all of the standard C math functions are imported into the global namespace. So doing something like `cos(...)` is not overload for `float`, `double`, `long double`, `std::complex<...>`, etc like you'd expect it to be in CXX. To use the actual STL math functions you MUST reference them through `::std`!

Why does this has that big of an effect? No idea. I'd imagine the age of GCC vs Clang could have something to do with it...?

I also noticed through a little bit more digging in the STL source code that that 'built-in' math functions are already handled for you, per type, in the STL (little bonus for us).

Super short version, only use math functions from `::std` and we should be fine. I will be opening up another ticket to make note of this and track any changes that need to be made to the code. All I can think of at the moment is a stray call to `sqrt(...)` in `lin::norm(...)`.

### DIFF
--- a/src/gnc_ode.cpp
+++ b/src/gnc_ode.cpp
@@ -223,8 +223,8 @@ int ode23(T ti, T tf, T const *yi, T *yf, unsigned int ne, T *bf, T h_min,
       // Calculate largest error
       delta_max = static_cast<T>(0);
       for (unsigned int i = 0; i < ne; i++) {
-        T delta = abs( ks[i] - kz[i] ) /
-            std::max(abs_tol, rel_tol * std::max(abs(ks[i]), abs(kz[i])));
+        T delta = std::abs( ks[i] - kz[i] ) /
+            std::max(abs_tol, rel_tol * std::max(std::abs(ks[i]), std::abs(kz[i])));
         if (delta > delta_max) delta_max = delta;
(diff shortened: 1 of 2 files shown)


tanishqaggarwal (approved):
Damn nice.

PR #77, opened 2019-11-08
Added Ode45 and Associated Unit Tests

### LINKED ISSUE #67: Variable Step Size Integrators
These will be useful in a couple use cases. For example, when we uplink a timestamped leader position, velocity to the follower it would be best to call a variable step size integrator to propagated the old timestamped value to the current time.

These function will live in `gnc_ode.hpp` and `gnc_ode.cpp`. They'll be titled `gnc::odeXX` where `XX` describes the orders of integration being compared.

We're looking to implement `ode23` and `ode45`.

kylekrol:
Closes #67.

Added `ode45` and made some optimizations (particularly I wasn't handling the recalculation of `k1` properly in either `ode23` or `ode45`).

Verified that all tests passed on a Teensy35 as well.

### DIFF
--- a/src/gnc_ode.cpp
+++ b/src/gnc_ode.cpp
@@ -14,7 +14,6 @@
 
 #include <algorithm>
 #include <cmath>
-#include <iostream>
 
 namespace gnc {
 
--- a/src/gnc_ode.cpp
+++ b/src/gnc_ode.cpp
@@ -14,6 +14,7 @@
 
 #include <algorithm>
 #include <cmath>
+#include <iostream>
 
 namespace gnc {
 
@@ -181,7 +182,6 @@ int ode23(T ti, T tf, T const *yi, T *yf, unsigned int ne, T *bf, T h_min,
   T *const kz = ks + ne;
 
   // Initialize local variables
-  bool calc_k1 = true;
   int err = ODE_ERR_OK;
   unsigned int iter = 0;
   T h_next = (tf - ti) / static_cast<T>(10);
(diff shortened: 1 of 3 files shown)


src/gnc_ode.cpp
```diff
@@ -14,6 +14,7 @@
 
 #include <algorithm>
 #include <cmath>
+#include <iostream>
```
tanishqaggarwal: Including this makes the code incompatible with Teensy

test/ode_test/test.cpp
```diff
@@ -107,10 +107,32 @@ void test_ode_ode23sho() {
   TEST_ASSERT_DOUBLE_WITHIN(2e-3, cos(7.5), yf[0]);
 }
 
+void test_ode_ode45sho() {
+  PAN_GNC_ODEXX_TEST_VARS(9);
+  int code;
```
tanishqaggarwal, line 112: For robustness, is it worth checking through many values for x in cos(x)? It'd be easy to do this with a for loop

src/gnc_ode.cpp
```diff
@@ -181,7 +182,6 @@ int ode23(T ti, T tf, T const *yi, T *yf, unsigned int ne, T *bf, T h_min,
   T *const kz = ks + ne;
 
   // Initialize local variables
-  bool calc_k1 = true;
```
tanishqaggarwal, line 184: Where did calc_k1 go? It still seems to be used below but I can't find the definition

tanishqaggarwal (changes requested):
My only woe is `<iostream>`

src/gnc_ode.cpp
tanishqaggarwal: Not "incompatible", but it just blows up the code size. You don't need it anyway, it seems

kylekrol: Lol yet no that was there from debugging and totally needs to be gone

src/gnc_ode.cpp
kylekrol: It's totally gone. I moved the initial `k1` calculation outside the loops.

test/ode_test/test.cpp
kylekrol: Good point. We could add more robust tests in the future. I will be working on the orbit propagator soon too - so much more testing to come!

tanishqaggarwal (approved):

Issue #78, opened 2019-11-08
Add Orbital State Estimation Function


kylekrol:
Depends on #65.

Add a CXX orbital state estimation function for GNC version 1.0.

In the far field, this function will provide position and velocity measurements for the leader and follower spacecrafts in ECI and ECEF. We will use a fixed step size integrator to propagate both orbits during each cycle of the flight computer.

If a timestamped position, velocity of the other satellite is received, it will be propagated forward to the present time with a variable step size integrator and replace the old position, velocity measurement of the old spacecraft.

In the near field, we will also have relative position, velocity in ECEF. Here we will add a simple state filter based on (most likely) Clohessy-Wiltshire dynamics to get a better relative velocity measurement. (@stewartaslan will open a separate issue for this filter - <link here>).

nhz2:
TODO, figure out what to do with float vs fixed RTK data

nhz2:
The estimator should also estimate the energy of both orbits.

nhz2:
The estimator should respond to / predict maneuvers.

nhz2:
Some issues with the current MATLAB orbit estimator implementation.
1. The orbit estimator can call the gravity function an unlimited number of times, this isn't possible on a teensy.
2. The estimator is numerically unstable.

To solve these I'm working on a redesigned orbit estimator, where the number of calls to the gravity function is controlled. Also in addition to the giant EKF the new filter will be multiple smaller filters running in parallel, each taking in different combinations of sensor readings/ ground data. That way if one EKF fails because of a bad sensor, the other filters can be used.

PR #81, opened 2019-11-08
Draft of Orbit Estimator Interface

kylekrol:
Relates to #78.

Opening this up for a sanity check.

@tanishqaggarwal I know we wanted to stay away from structs, but this is turning into a ridiculous amount of data that may or may not exist so feel like wrapping stuff up in structs in unfortunately warranted.

If anyone has any questions about the interface, how optional data will be handled, et cetera just let me know. I haven't fully documented the `estimate_orbit` function yet.

### DIFF
--- a/include/gnc_orbit_estimation.hpp
+++ b/include/gnc_orbit_estimation.hpp
@@ -1,0 +1,94 @@
+//
+// include/gnc_orbit_estimation.hpp
+// PSim
+//
+// Contributors:
+//   Kyle Krol  kpk63@cornell.edu
+//
+// Pathfinder for Autonomous Navigation
+// Space Systems Design Studio
+// Cornell Univeristy
+//
+
+#ifndef PAN_PSIM_INCLUDE_GNC_ESTIMATION_HPP_
+#define PAN_PSIM_INCLUDE_GNC_ESTIMATION_HPP_
+
+#include "lin.hpp"
+
+namespace gnc {
+
+/** @struct OrbitState
+ *  Minimal representation of a spacecraft's orbital state. It contains
+ *  timestamped position and velocity vectors in ECI. The time should be seconds
+ *  since the 'PAN epoch'. */
+struct OrbitState {
+  double t;
+  lin::Vector3d r_eci;
+  lin::Vector3d v_eci;
+};
+
+/** @struct OrbitStateEstimate
+ *  Serves as an output from the orbit estimation function. It's an overdefined
+ *  state that includes timestamped position and velocity vector in ECI and
+ *  ECEF. The time should be seconds since the 'PAN epoch'. */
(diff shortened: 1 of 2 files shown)


include/gnc_orbit_estimation.hpp
```diff
@@ -0,0 +1,94 @@
...
+ *                         unknown).
+ *  @param[in]  gps1       GPS data from this satellite (nullptr if unknown).
+ *  @param[in]  gps2       GPS data from that satellite (nullptr if unknown).
+ *  @param[in]  t          Desired timestamp for the output estimate.
+ *  @param[out] state1_est Destination for the state estimate of this satellite
+ *                         (set to nullptr if undetermined).
+ *  @param[out] state2_est Destination for the state estimate of that satellite
+ *                         (set to nullptr if undetermined). */
+void estimate_orbit(OrbitState const *state1, OrbitState const *state2,
+    GpsData const *gps1, GpsData const *gps2, double t,
```
tanishqaggarwal, line 65: What happens if the GPS measurement is from a time a little ways away in the past?

include/gnc_orbit_estimation.hpp
```diff
@@ -0,0 +1,94 @@
...
+ *  ECEF, and relative RTK position in ECEF. */
+struct GpsRtkData {
+  double t;
+  lin::Vector3d r_ecef;
+  lin::Vector3d v_ecef;
+  lin::Vector3d dr_ecef;
+};
+
+/** @struct RtkOrbitEstimatorState */
+struct RtkOrbitEstimatorState { };
```
tanishqaggarwal, line 79: What do you plan to go into this struct?

tanishqaggarwal (commented):
I think the structs are totally fine actually, I don't see how you could avoid them. I just have a couple of questions.

include/gnc_orbit_estimation.hpp
kylekrol: Yep, so for the `gps1` (i.e. the satellite running the algorithm) the GPS data should always be up to date and it's time value should be the same as `t` even. So no concerns there. For the other satellites data, however, you are right that it could be from a really long time ago. In that case, the algorithm will check for that and if need be propagate the orbit forward until the present and do whatever it's gonna do with that.

include/gnc_orbit_estimation.hpp
kylekrol: This will hold some internal state information. Like extra variables that may need to persist across calls for the sake of the filter. We'll store a timestamp in there too so if the function is called with data that's "too old" we can reset internally.

kylekrol:
Sounds good - okay and @nhz2 I'm going to rework some of this stuff so that any input that aren't available or valid will be set to NaN. Same with outputs!

Issue #66, opened 2019-11-04
Add Attitude Estimation Function (1.0)


kylekrol:
Add a CXX attitude state estimation function for GNC version 1.0.

This function will provide an attitude estimate whenever an ECI position, time, magnetic field reading, and sun vector reading are provided (i.e. pretty much whenever a sun vector is available). It will rely on simple triad and return a quaternion specifying the relationship between the body frame and ECI.

No attitude state filtering will be included.

kylekrol:
@nhz2 @stewartaslan Let me know if anything here isn't what we discussed!

PR #83, opened 2019-11-09
Draft of Attitude Estimator Function

kylekrol:
Relates to #66.

Here is what I would consider a fairly finalized draft of the GNC v1.0 attitude estimation function. Let me know what we all think!

I think the struct design paradigm here is very beneficial for a couple reasons:
 1. All data not specified will come back as `NaN`. This make it easy to handle optional data and very clear when some value wasn't calculate properly. I will work on adding limited `NaN` support to lin so the amount of boilerplate code can be reduced.
 2. Adding extra input variables won't break FSW builds! The variable just won't be set until FSW is updated. This will greatly reduce the amount of versioning we need to do.
 3. It prevents us from having a million input arguments.
 4. We don't have to pass any `nullptr`'s around or have pointers to pointers.

Let me know your thoughts! I will restructure the orbit estimator to mirror the structure presented here a little more closely.

The standard usage pattern I had in mind would be something like this:

    AttitudeEstimatorState state;
    AttitudeEstimate estimate;

    repeat {

      AttitudeEstimatorData data;  // Initial populate everything win NaN
      data.t = ...  // Assign all known data
      ...

      gnc::estimate_attitude(state, data, estimate);

      ... = estimate.q_body_eci;  // Extract all outputs and check for NaN
      ...
    
    }

### DIFF
--- a/include/gnc_attitude_estimation.hpp
+++ b/include/gnc_attitude_estimation.hpp
@@ -1,0 +1,62 @@
+//
+// include/gnc_attitude_estimation.hpp
+// PSim
+//
+// Contributors:
+//   Kyle Krol  kpk63@cornell.edu
+//
+// Pathfinder for Autonomous Navigation
+// Space Systems Design Studio
+// Cornell Univeristy
+//
+
+#ifndef PAN_PSIM_INCLUDE_GNC_ATTITUDE_ESTIMATION_HPP_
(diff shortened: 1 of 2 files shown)


kylekrol:
This was supposed to be a draft for now but I guess it doesn't really matter.

PR #84, opened 2019-11-10
Added Attitude Estimation Function with Limited Unit Tests

### LINKED ISSUE #66: Add Attitude Estimation Function (1.0)

kylekrol:
Closes #66.

Here is a much more finalized initial implementation of the attitude estimation function.

There are a couple PSAs that also need to go out with this PR:

 * It was discovered that including `#include <math.h>` after including `#include <cmath>` will break the `std::isnan(...)` function because there is a `isnan` macro defined in `math.h` and undefined in `cmath`.
 * `#include <unity.h>` for unit testing will include `math.h` by default and define the `isnan` and `isinf` macros regardless. Therefore it needs to be included like this:

        #define UNITY_EXCLUDE_MATH_H
        #include <unity.h>
        #undef isinf
        #undef isnan

More unit tests will obviously need to be added. However, I wanted to keep these PR's of manageable size.

The `estimate_attitude` function and associated structs should be used in this sort of fashion:

    gnc::AttitudeEstimatorState state;
    gnc::AttitudeEstimatorData data;
    gnc::AttitudeEstimate estimate;

    repeat {

      data = gnc::AttitudeEstimatorData();
      data.t = ...  // Populate all state fields that are valid (otherwise leave them defaulted to NaN)
      ...

      gnc::estimate_attitude(state, data, estimate);

      if (!std::isnan(estimate.q_body_eci(0)))  // Copy in all outputs that aren't NaN
        ... = estimate.q_body_eci
      ...

    }


include/gnc_containers.hpp
```diff
@@ -0,0 +1,60 @@
+//
```
tanishqaggarwal, line 1: I already had this on the CommonSoftware side. The testing is a little lacking on my version though

tanishqaggarwal: I'm working on porting my version to work with dynamically allocated memory in a stack structure. We need this for storing downlink packets.

tanishqaggarwal: Let's talk integration at some point? It's unfortunate that we separately developed this.

src/gnc_attitude_estimation.cpp
```diff
@@ -0,0 +1,122 @@
...
+ *  Approximates the angular rate of a body given a quaternion q_m measured at a
+ *  time t and a quaternion q_n measured at a time t + dt. */
+static void estimate_w_body(double dt, lin::Vector4f const &q_m,
+    lin::Vector4f const &q_n, lin::Vector3f &w) {
+  // Find the 'difference' between the two quaternions
+  lin::Vector4f q_n_conj, dq;
+  utl::quat_conj(q_n, q_n_conj);
+  utl::quat_cross_mult(q_m, q_n_conj, dq);
+  // Approximate w_body
+  w = lin::ref<3, 1>((2.0f / static_cast<float>(dt)) * dq, 0, 0);
```
tanishqaggarwal: I'm confused as to how 2/dt * dq yields a scalar value.

test/containers_test/test.cpp
```diff
@@ -0,0 +1,58 @@
...
+  TEST_ASSERT_FALSE(buffer.empty());
+  TEST_ASSERT_TRUE(buffer.full());
+  TEST_ASSERT_EQUAL_FLOAT(buffer[0], 1.0f);
+  TEST_ASSERT_EQUAL_FLOAT(buffer[1], 2.0f);
+  TEST_ASSERT_EQUAL_FLOAT(buffer[2], 3.0f);
+}
+
+void test_circular_buffer() {
+  gnc::CircularBuffer<float, 3> buffer;
+  for (unsigned int i = 0; i < 4; i++) { // Test many starting offsets
```
tanishqaggarwal: What are "offsets" and how are you testing them here?

test/attitude_estimation_test/test.cpp
```diff
@@ -0,0 +1,30 @@
...
+// Space Systems Design Studio
+// Cornell Univeristy
+//
+
+#include "../test.hpp"
+
+#include <gnc_attitude_estimation.hpp>
+#include <cmath>
+
+void test_nan_intputs() {
```
tanishqaggarwal: Typo: intputs

src/gnc_attitude_estimation.cpp
```diff
@@ -0,0 +1,122 @@
...
+}
+
+void estimate_attitude(AttitudeEstimatorState &state,
+    AttitudeEstimatorData const &data, AttitudeEstimate &estimate) {
+
+  // Clear the state if it's outdated (more than ~5 control cycles old)
+  if (data.t - state.t > 0.5) state = AttitudeEstimatorState();
```
tanishqaggarwal, line 102: Why do we need to check for this? This function will be called on every control cycle. Is the concern is that the data might get stale due to ADCS getting disconnected?

src/gnc_attitude_estimation.cpp
```diff
@@ -0,0 +1,122 @@
...
+    lin::Vector4f const &q_n, lin::Vector3f &w) {
+  // Find the 'difference' between the two quaternions
+  lin::Vector4f q_n_conj, dq;
+  utl::quat_conj(q_n, q_n_conj);
+  utl::quat_cross_mult(q_m, q_n_conj, dq);
+  // Approximate w_body
+  w = lin::ref<3, 1>((2.0f / static_cast<float>(dt)) * dq, 0, 0);
+}
+
+void estimate_attitude(AttitudeEstimatorState &state,
```
tanishqaggarwal, line 98: Why is there a difference between EstimatorState and Estimate? They seem to have the same variables except for w_buffer. Why's w_buffer needed?

src/gnc_attitude_estimation.cpp
```diff
@@ -0,0 +1,122 @@
...
+ *  quaternion q_body_eci. */
+static int estimate_q_body_eci(double t, lin::Vector3d const &r_ecef,
+    lin::Vector3f const &s_body, lin::Vector3f const &b_body, lin::Vector4f &q_body_eci) {
+  // Setup the rest of the triad inputs
+  lin::Vector3f s_eci, b_eci, r_ecef_f({
+    static_cast<float>(r_ecef(0)),
+    static_cast<float>(r_ecef(1)),
+    static_cast<float>(r_ecef(2))
+  });
+  env::sun_vector(t, s_eci);
```
tanishqaggarwal, line 65: Can you add a comment here explaining what these two env:: functions do?

include/gnc_containers.hpp
```diff
@@ -0,0 +1,60 @@
...
+  constexpr CircularBuffer(CircularBuffer<T, N> const &) = default;
+  constexpr CircularBuffer(CircularBuffer<T, N> &&) = default;
+  constexpr CircularBuffer<T, N> &operator=(CircularBuffer<T, N> const &) = default;
+  constexpr CircularBuffer<T, N> &operator=(CircularBuffer<T, N> &&) = default;
+  /** @returns True if the buffer contains the maximum number of elements. */
+  constexpr bool full() const;
+  /** @returns True if the buffer contains no elements. */
+  constexpr bool empty() const;
+  /** @returns Number of elements currently in the buffer. */
+  constexpr std::size_t size() const;
```
tanishqaggarwal, line 39: size_t is available in `<cstddef>` without the need for std::

include/inl/gnc_containers.inl
```diff
@@ -0,0 +1,64 @@
...
+  return n;
+}
+
+template <typename T, std::size_t N>
+constexpr std::size_t CircularBuffer<T, N>::max_size() const {
+  return N;
+}
+
+template <typename T, std::size_t N>
+constexpr T const &CircularBuffer<T, N>::operator[](std::size_t i) const {
```
tanishqaggarwal, line 42: This implementation is missing a pop() function. It's not necessary for your application but just thought I'd point that out as a generally needed feature of ring buffers (we definitely need it on the flight software side)

include/gnc_attitude_estimation.hpp
```diff
@@ -0,0 +1,72 @@
...
+  /** Timestamp for the rest of the data in this struct and the time for which
+   *  an attitude estimate will be calculated. This should be in seconds since
+   *  the 'PAN epoch'. */
+  double t;
+  /** Position of the satellite in ECEF (units m). */
+  lin::Vector3d r_ecef;
+  /** Magnetic field vector in the body frame of the spacecraft (units T). */
+  lin::Vector3f b_body;
+  /** Sun vector in the body frame of the spacecraft (unit vector). */
+  lin::Vector3f s_body;
```
tanishqaggarwal, line 45: Don't we also need sun sensor vector validity here?

tanishqaggarwal (changes requested):
Left a few comments and questions. Interface looks good; I'll start modifying the draft PR on Flight Software to match this.

This seems to be missing any use of gyro data, magnetometer data, wheel speeds, or magnetorquer command data. Is that in the works, or is this the final interface?

include/gnc_attitude_estimation.hpp
kylekrol: Sun vector validity (and validity for all measurements in fact) will just be conveyed by whether or not a field in `nan`. That's why all the constructors default all values to `nan` or in the case of the buffer empty.

include/inl/gnc_containers.inl
kylekrol: Ah okay yeah I could add that in pretty easily -> just bump up the `s` member variable by one. `peek` and `pop` work?

include/gnc_containers.hpp
kylekrol: Didn't know that was a thing - although forgot I didn't use `array` because of `constexpr` issues so I gotta change some of the include stuff up anyway. Seems a little weird that you don't need `std::` though if it's part of the STL. That `size_t` in particular is actually probably from `C` and we may want to avoid using it for consistency in `cxx` code. I've been having issues including both `math.h` and `cmath` due to naming conflicts and such.

src/gnc_attitude_estimation.cpp
kylekrol: I can. Although if you just hover over them should pull up the documentation?

test/containers_test/test.cpp
kylekrol: Different starting positions within the buffer (see member variable `s` I think).

src/gnc_attitude_estimation.cpp
kylekrol: The first argument to `lin::ref<...>(...)`, `(2.0f / static_cast<float>(dt)) * dq`, actually yields a `lin::internal::ScaleMultiple<...>` which is essentially a lazily evaluated `lin::Vector4f`.  It gets passed to the reference function and only the first three components actually calculated for the assignment to `w`.

src/gnc_attitude_estimation.cpp
kylekrol: `AttitudeEstimatorState` holds internal variables that the estimate function operates on. We stored old `w` and `q` measurements in here to facilitate the calculation of angular rate after more than one successful call to `estimate_attitude`.

The `AttitudeEstimate` variable will actually contain stuff that FSW will read out. In short, FSW should never directly manipulate `AttitudeEstimateState`, should write to `AttitudeEstimateData`, and should read from `AttitudeEstimate`.

src/gnc_attitude_estimation.cpp
kylekrol: Yes because the last saved quaternion in the internal state could be from like before eclipse. In that case we surely don't want to try to use it to calculate angular rate. There are plenty of scenarios where this function fails other than just having ADCS not responding.

kylekrol:
This is the final interface for what we are calling GNC v1.0. In future versions when we hopefully have a state filter and everything we will add all of those inputs to the data struct and then add more outputs in the estimate struct. I think @nhz2 would be a good person to talk to about what this could end up looking like.

This is part of the reason for using the struct interface as opposed to actual arguments. This way we can change things on the PSim side of things (i.e. add struct members) but it won't break the FSW builds. Obviously the control tasks will have to work off of releases in some capacity however in tandem with GNC.

include/gnc_containers.hpp
tanishqaggarwal: `size_t` is enough of a primitive type that I just went for using it in Flight Software; it's all over the place.` <cstddef>` wouldn't interfere with math, I think, so it's available if you want to make the code more consistent with FSW. Up to you.

src/gnc_attitude_estimation.cpp
tanishqaggarwal: I think the purpose of code should always be as readily apparent as possible.

include/gnc_containers.hpp
tanishqaggarwal: Upon further discussion with FSW we actually don't need the circular buffers for telemetry, so I'm getting rid of them. Yours is now the canonical implementation (which is good, since it actually works! :P)

src/gnc_attitude_estimation.cpp
tanishqaggarwal: So if I understand correctly, w is being set to the first three components of the quaternion 2/dt * dq?

src/gnc_attitude_estimation.cpp
tanishqaggarwal: Ah, forgot about eclipse. Thanks.

src/gnc_attitude_estimation.cpp
tanishqaggarwal: Got it, `State` is an intermediate.

include/inl/gnc_containers.inl
tanishqaggarwal: Sure! Also we don't need ring buffers on FSW anymore, so it's a totally optional feature

include/gnc_attitude_estimation.hpp
tanishqaggarwal: That's clever!

include/gnc_containers.hpp
kylekrol: What do you mean to work with dynamically allocated memory in a stack structure?

include/inl/gnc_containers.inl
kylekrol: Ah okay gotcha - probs just leave it for now then

include/gnc_containers.hpp
tanishqaggarwal: I mean that the size of the buffer is computed at runtime, but once it's computed the buffer size doesn't change.

For downlinking we originally wanted fixed-size FIFO buffers, which is what I mean by stack structure.

test/test_all/test.hpp
```diff
@@ -0,0 +1,27 @@
...
+// Cornell Univeristy
+//
+
+#ifndef PAN_PSIM_TEST_TEST_ALL_TEST_HPP_
+#define PAN_PSIM_TEST_TEST_ALL_TEST_HPP_
+
+#define UNITY_EXCLUDE_MATH_H
+#include <unity.h>
+#undef isnan
+#undef isinf
```
kylekrol, line 19: If `unity.h` is not included in this fashion and you are compiling GNC source along with other unit tests, the compilation will likely fail due to the `isnan` macro! Just a heads up.

kylekrol:
Sorry for the big rework, but there was no point in doing individual source builds for each of the unit tests so I smashed them all into one!

tanishqaggarwal (dismissed):
I'm not a fan of combining all the tests under one folder, since it makes the compile-test-run cycle a lot longer if any individual function breaks due to any coming improvements. But it doesn't affect functionality, so I'm approving.

tanishqaggarwal:
> I'm not a fan of combining all the tests under one folder, since it makes the compile-test-run cycle a lot longer if any individual function breaks due to any coming improvements. But it doesn't affect functionality, so I'm approving.

I guess you can also just comment out tests in the `test_all` file, nevermind.

kylekrol:
> I'm not a fan of combining all the tests under one folder, since it makes the compile-test-run cycle a lot longer if any individual function breaks due to any coming improvements. But it doesn't affect functionality, so I'm approving.

Didn't think about that but yeah commenting out should be good!

kylekrol:
@nhz2 I know we talked about this over the phone but I'll wait till you take a look. If you could glance at the `w_body` calculation especially that'd be awesome!

include/gnc_attitude_estimation.hpp
```diff
@@ -0,0 +1,72 @@
...
+};
+
+/** @struct AttitudeEstimate
+ *  Essentially serves as the outputs to the estimate_attitude function. See
+ *  member documentation for more details. */
+struct AttitudeEstimate {
+  /** Quaternion transforming from the body frame to ECI. */
+  lin::Vector4f q_body_eci;
+  /** Angular rate of the spacecraft in the body frame. */
+  lin::Vector3f w_body;
```
nhz2, line 57: Can you add an estimate of the magnitude of the angular momentum, which would be used to determine whether to go to detumble mode?

nhz2: The advantage to having this as a possible separate output, is that the magnitude of the angular momentum is expected to change very slowly, so when we implement filters it can have a very long time constant.

src/gnc_attitude_estimation.cpp
```diff
@@ -0,0 +1,123 @@
...
+  // field vectors in ECI)
+  lin::Vector3f s_eci, b_eci, r_ecef_f({
+    static_cast<float>(r_ecef(0)),
+    static_cast<float>(r_ecef(1)),
+    static_cast<float>(r_ecef(2))
+  });
+  env::sun_vector(t, s_eci);
+  env::magnetic_field(t, r_ecef_f, b_eci);
+  // Perform triad and return the result code (result is written to q_body_eci)
+  return utl::triad(s_eci, b_eci, s_body, b_body, q_body_eci);
```
nhz2: All the vectors going into triad need to be unit vectors, so try to normalize `b_body` and `b_eci`, fail the estimation if either are too small (under the noise/bias level of the magnetometer).

nhz2 (changes requested):

src/gnc_attitude_estimation.cpp
```diff
@@ -0,0 +1,123 @@
...
+ *  Approximates the angular rate of a body given a quaternion q_m measured at a
+ *  time t and a quaternion q_n measured at a time t + dt. */
+static void estimate_w_body(double dt, lin::Vector4f const &q_m,
+    lin::Vector4f const &q_n, lin::Vector3f &w) {
+  // Find the 'difference' between the two quaternions
+  lin::Vector4f q_n_conj, dq;
+  utl::quat_conj(q_n, q_n_conj);
+  utl::quat_cross_mult(q_m, q_n_conj, dq);
+  // Approximate w_body
+  w = lin::ref<3, 1>((2.0f / static_cast<float>(dt)) * dq, 0, 0);
```
nhz2: Check the sign of `dq[4]`, if it is negative, negate `w` as well. This is because this calc assumes `dq[4]` is positive, so if it isn't just negate all of dq.

include/gnc_attitude_estimation.hpp
kylekrol: Would that be a separate filter input output then? It'd seem weird to have the magnitude of omega not match with the magnitude of the omega vector.

src/gnc_attitude_estimation.cpp
kylekrol: Very good point! I totally missed that!

include/gnc_attitude_estimation.hpp
kylekrol: Discussed offline that FSW will calculate the magnitude of `w_body` externally.

src/gnc_attitude_estimation.cpp
```diff
@@ -45,33 +46,53 @@ AttitudeEstimate::AttitudeEstimate() {
...
     static_cast<float>(r_ecef(0)),
     static_cast<float>(r_ecef(1)),
     static_cast<float>(r_ecef(2))
   });
   env::sun_vector(t, s_eci);
   env::magnetic_field(t, r_ecef_f, b_eci);
+
+  // Verify magnetic field vectors have are large enough
+  constexpr static float threshhold = constant::b_noise_floor_f * constant::b_noise_floor_f;
+  if (lin::fro(b_eci) <= threshhold || lin::fro(b_body) <= threshhold)
```
tanishqaggarwal: threshhold is spelled with one h

src/gnc_attitude_estimation.cpp
```diff
@@ -45,33 +46,53 @@ AttitudeEstimate::AttitudeEstimate() {
...
+ *     floor of the magnetometer.
+ *   - Magnitude of the measured magnetic field is above the noise floor of the
+ *     magnetometer. */
 static int estimate_q_body_eci(double t, lin::Vector3d const &r_ecef,
     lin::Vector3f const &s_body, lin::Vector3f const &b_body, lin::Vector4f &q_body_eci) {
+
   // Setup the rest of the triad inputs (including expected sun and magnetic
   // field vectors in ECI)
-  lin::Vector3f s_eci, b_eci, r_ecef_f({
+  lin::Vector3f s_eci, b_eci, b_u_body, r_ecef_f({
```
tanishqaggarwal, line 60: What's b_u?

tanishqaggarwal (dismissed):

src/gnc_attitude_estimation.cpp
kylekrol: Yep oops... Getting that now

src/gnc_attitude_estimation.cpp
kylekrol: And just a unit vector (`b_body` is `const` so I had to make another variable for it). It's assigned like 5 lines down.

tanishqaggarwal (approved):

nhz2 (approved):

PR #88, opened 2019-11-25
Plots

nhz2:


MATLAB/test/test_get_truth.m
```diff
@@ -38,4 +40,6 @@
 get_truth('inclination angle',d);
 get_truth('right ascension of the ascending node',d);
 get_truth('argument of perigee',d);
-get_truth('true anamoly',d);
\ No newline at end of file
+get_truth('true anamoly',d);
```
tanishqaggarwal: anomaly

tanishqaggarwal:
When are the plot functions actually called? Is the call in "main"?

nhz2:
The plot functions can be called after `main` runs, because `main` saves the trajectory in `main_state_trajectory`. For example you should be able to run, `plot_get_truth(main_state_trajectory,'wheel rate leader (rad/s) body','l')` and get a plot like this

wheel_speed

kylekrol (approved):

PR #90, opened 2019-11-26
Added single-satellite simulation to usb_console

tanishqaggarwal:
Be sure to check out configs to see the config changes.

Also, some bad news: currently everything seems to be running at ~0.6x real time, despite no state exchange happening between MATLAB and the Teensies. I'll investigate this when I find the time

### DIFF
--- a/python/usb_console/cmdprompt.py
+++ b/python/usb_console/cmdprompt.py
@@ -4,7 +4,6 @@ except ImportError:
     # We're on Windows, so readline doesn't exist
     pass
 from cmd import Cmd
-import random
 import timeit
 
 class StateCmdPrompt(Cmd):
@@ -24,10 +23,12 @@ class StateCmdPrompt(Cmd):
 
         # By default, if it's available, set the prompt to be commanding the Flight Controller.
         try:
-            self.cmded_device = self.devices['FlightControllerLeader']
-
+            self.cmded_device = self.devices['FlightController']
         except KeyError:
-            self.cmded_device = random.choice(list(self.devices.values()))
+            try:
+                self.cmded_device = self.devices['FlightControllerLeader']
+            except KeyError:
+                self.cmded_device = list(self.devices.values())[0]
 
         Cmd.__init__(self)
 
--- a/python/usb_console/run_simulation.py
+++ b/python/usb_console/run_simulation.py
@@ -5,7 +5,7 @@ from state_session import StateSession
 from radio_session import RadioSession
 from cmdprompt import StateCmdPrompt
 from data_consumers import Logger, Datastore
-from simulation import Simulation
+from simulation import Simulation, SingleSimulation
 import json, sys, os, tempfile, time
 
 try:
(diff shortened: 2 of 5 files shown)


python/usb_console/cmdprompt.py
```diff
@@ -24,10 +23,12 @@ def __init__(self, devices, radios, sim, exit_fn):
 
         # By default, if it's available, set the prompt to be commanding the Flight Controller.
         try:
-            self.cmded_device = self.devices['FlightControllerLeader']
-
+            self.cmded_device = self.devices['FlightController']
         except KeyError:
-            self.cmded_device = random.choice(list(self.devices.values()))
```
tanishqaggarwal, line 30: randomly choosing a device was so stupid

python/usb_console/run_simulation.py
```diff
@@ -117,7 +118,7 @@ def set_up_radios(self):
                 radio_data_name = radio_connected_device + "_radio"
                 radio_datastore = Datastore(radio_data_name, self.simulation_run_dir)
                 radio_logger = Logger(radio_data_name, self.simulation_run_dir)
-                radio_session = RadioSession(radio_connected_device, radio_datastore, radio_logger, self.radio_keys_config)
+                radio_session = RadioSession(radio_name, radio_datastore, radio_logger, self.radio_keys_config)
```
tanishqaggarwal, line 121: This is to fix name clashes between the "FlightController" Teensy and the "FlightController" radio when commanding state via the command prompt

python/usb_console/simulation.py
```diff
@@ -136,3 +138,81 @@ def stop(self, data_dir):
 
         with open(data_dir + "/simulation_log.txt", "w") as fp:
             fp.write(self.log)
+
+class SingleSimulation(Simulation):
```
tanishqaggarwal: @kkrol27 this kind of abstraction is how the company we love built our most favorite, gnarly test product

tanishqaggarwal:
@nhz2 As Kyle predicted, this solution ended up being pretty hacky. Would it be a good idea for MATLAB to more natively support a single-sat simulation?

nhz2:
Can't you just sim the leader FC using `update_FC_state` in matlab and only have the follower in HITL/SITL?

python/usb_console/simulation.py
kylekrol: Yep looks like it haha

kylekrol:
> Can't you just sim the leader FC using `update_FC_state` in matlab and only have the follower in HITL/SITL?

Just to emphasize this, the way I sorta had seen this going down was, as Nathan suggested, have the MATLAB `update_FC_state` act as a minimum working example of a FC stepping through a completely nominal mission.

That being said, it may be advantageous to take a single sat sim even further. In such a case we could have the second spacecraft simple be a dummy target. It would always have it's docking end pointed directly at the follower spacecraft and we wouldn't need to sim any FC behavior, actuators, sensors, or physics (except for a point mass orbit propagator of course). I'm not sure how possible this is now, but I could take some time to defamiliarize myself with the MATLAB code we have now to see if this is possible.

Let me know if anyone has additional thoughts, comments, concerns.

tanishqaggarwal:
I think that's exactly what's going on right now @kkrol27

kylekrol (approved):

Issue #82, opened 2019-11-08
Add Attitude Controller Function


kylekrol:
Add a CXX attitude controller function for GNC version 1.0.

There should be two main controllers: detumble and pointing.

The former will use @nhz2's b-dot controller to generate actuation commands and the latter will take in a current attitude estimate and a desired attitude.

PR #91, opened 2019-11-26
Added Attitude Detumble Controller

kylekrol:
Relates to #82.

Completed the detumble attitude controller. The pointing attitude controller should be coming next! It follows the same design pattern as the attitude estimator.

One issue that is starting to crop up is the duplication of constants between common software and psim (in this PR it's the `gnc::constants::max_mtr_moment`). I think it'll be easiest to refactor ADCS Software and the ADCS driver down the line - otherwise the dependency web will start getting bit crazy. However, I won't push for this change until GNC stuff is in a more completed state (i.e. like the version 1.0 release).

### DIFF
--- a/src/gnc_attitude_controller.cpp
+++ b/src/gnc_attitude_controller.cpp
@@ -59,6 +59,7 @@ void control_detumble(DetumbleControllerState &state,
       constant::max_mtr_moment_f * (db(1) > 0.0f ? -1.0f : 1.0f),
       constant::max_mtr_moment_f * (db(2) > 0.0f ? -1.0f : 1.0f)
     };
+    state.b_body_buffer.clear();
   }
 
   // Assign the most recently calculated magetorquer command
--- a/src/gnc_attitude_controller.cpp
+++ b/src/gnc_attitude_controller.cpp
@@ -1,0 +1,72 @@
+//
+// src/gnc_attitude_controller.hpp
+// PSim
+//
+// Contributors:
+//   Kyle Krol  kpk63@cornell.edu
+//
+// Pathfinder for Autonomous Navigation
+// Space Systems Design Studio
+// Cornell Univeristy
+//
+
+#include <gnc_attitude_controller.hpp>
+#include <gnc_constants.hpp>
+
+#include <limits>
+
+static_assert(std::numeric_limits<double>::has_quiet_NaN,
+    "GNC code requires quiet NaN's to be available.");
+static_assert(std::numeric_limits<float>::has_quiet_NaN,
+    "GNC code requires quiet NaN's to be available.");
+
+namespace gnc {
+
+DetumbleControllerState::DetumbleControllerState() {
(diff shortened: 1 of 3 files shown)


kylekrol:
Ah right sorry @shihaocao, I forgot you were working on the flight computer side of things! I'll add you to future FSW related GNC PRs.

tanishqaggarwal (dismissed):
Looks good to me!

kylekrol:
I'm gonna wait to merge until @nhz2 takes a look as well. I'll start working off of this branch to add the pointing controller though.

src/gnc_attitude_controller.cpp
```diff
@@ -0,0 +1,72 @@
...
+  lin::Vector3f db = (data.b_body - state.b_body) / (data.t - state.t);
+  state.db_buffer.push(db);
+  db = lin::zeros<float, 3, 1>();
+  for (std::size_t i = 0; i < state.db_buffer.size(); i++)
+    db = db + state.db_buffer[i];
+  db = db / static_cast<float>(state.db_buffer.size());
+
+  // Run the bang-bang controller
+  actuation.mtr_body_cmd = {
+    constant::max_mtr_moment_f * (db(0) > 0.0f ? -1.0f : 1.0f),
```
nhz2: If the magrod command changes that destroys the bdot estimate, because, the magnetic field bias from the magrods change. Thats why I had the command only change every 10 adcs cycles.

src/gnc_attitude_controller.cpp
```diff
@@ -0,0 +1,72 @@
...
+    state.b_body = data.b_body;
+  }
+
+  // Calculate db and calculate its moving average
+  lin::Vector3f db = (data.b_body - state.b_body) / (data.t - state.t);
+  state.db_buffer.push(db);
+  db = lin::zeros<float, 3, 1>();
+  for (std::size_t i = 0; i < state.db_buffer.size(); i++)
+    db = db + state.db_buffer[i];
+  db = db / static_cast<float>(state.db_buffer.size());
```
nhz2: db should only average over a period of time where the magrod command hasn't changed

nhz2 (changes requested):
db should only average over a period of time where the magrod command hasn't changed.
Also, the bdot controller shouldn't take time as an input, because it has to run before we have valid GPS. Lastly, since its a bang bang controller, don't worry about dividing by dt, or the averaging constant.

kylekrol:
Yeah nope totally forgot about not having GPS or anything like that. I've updated it to be far simpler. Now it waits until it's been given ten magnetic field readings and using the 5 most recent readings to calculate a `db` to feed into the bang bang controller. It then clears the buffer and repeats. On loops where it doesn't calculated a new actuation, it simply returns the previously calculated MTR command.

I had it look at the most recent half of the buffer to avoid any weird transients from the updated actuation affecting our `db` calculation.

nhz2 (approved):

PR #92, opened 2019-11-27
Adcs controllers and main pointing strategy

nhz2:


MATLAB/config.m
```diff
@@ -70,6 +70,7 @@
 % Torque on fuel/difference in angular rates in eci Nm/(rad/s).
 const.ATTITUDE_PD_KP = 75e-4; % imported from simulink
 const.ATTITUDE_PD_KD = 32.5e-4; % imported from simulink
+const.ATTITUDE_PD_derivative_buffer_size=1;
```
kylekrol: Will we need this for the flight algorithm you think? We're already buffering `w_body` in the estimation step.

MATLAB/detumbler.m
```diff
@@ -0,0 +1,16 @@
...
+        state.old_magnetic_field= magnetometer_body;
+    case 9
+        %calculate finite difference and bang bang controller
+        deltaB= magnetometer_body-state.old_magnetic_field;
+        state.magrod_moment_cmd= -sign(deltaB)*const.MAXMOMENT;
+end
+state.counter= mod(state.counter + 1,10);
+magrod_moment_cmd= state.magrod_moment_cmd;
+end
+
```
kylekrol: I'm glad you broke these out into separate functions. I was going to do that in order to be able to stub in some cxx stuff!

MATLAB/recursive_least_squares.m
```diff
@@ -0,0 +1,19 @@
...
+    B_measured= B_true+B_bias_true+0.02*randn(3,1);
+    SdotB_true= dot(B_true,S);
+    SdotB_measured= dot(B_measured,S);
+    SdotB_bias= SdotB_measured-SdotB_true;
+    P=P-P*S'*S*P/(1+S*P*S');
+    B_bias_est= B_bias_est + P*S'*(SdotB_bias-dot(S,B_bias_est));
+    error(i)=norm(B_bias_est-B_bias_true);
+end
+
+plot(error)
```
kylekrol: What exactly is going on here and what are we using it for? Sorry I'm out of the loop on this!

MATLAB/sensor_reading.m
```diff
@@ -48,9 +48,15 @@
...
 sensor_readings.position_ecef= position_ecef;
-%TODO get velocity in ECEF
+sensor_readings.velocity_ecef= utl_rotateframe(quat_ecef_eci, true_state.velocity_eci)-cross(rate_ecef,position_ecef);
 
 
+%in reality this is not how cd cps works
+target_position_eci= other_satellite_state.dynamics.position_eci;
+sensor_readings.target_position_ecef= utl_rotateframe(quat_ecef_eci,target_position_eci);
+target_velocity_eci= other_satellite_state.dynamics.velocity_eci;
+sensor_readings.target_velocity_ecef= utl_rotateframe(quat_ecef_eci, target_velocity_eci)-cross(rate_ecef,sensor_readings.target_position_ecef);
```
kylekrol, line 58: Does CD GPS actually give us a velocity reading or were we going to calculate that ourselves?

MATLAB/sensor_reading.m
```diff
@@ -102,7 +108,7 @@
             Y(j) = voltages(i);
         end
     end
-    if j >= 4  % TODO : Perhaps play with this parameter
+    if j >= 3  % TODO : Perhaps play with this parameter
```
kylekrol, line 111: What was the motivation for changing this value? I'm not saying I did an extensive dig into why I set it at 4 and not 3, but I was more comfortable having the system over-constrained. Especially when the sun vector is towards the docking end of the satellite (otherwise we'd always have 4+ sensors in view of the sun anyway).

MATLAB/update_FC_state.m
```diff
@@ -12,91 +12,149 @@
...
+estimator.tumbling= ~(norm(estimator.angular_momentum_body)<const.MAXWHEELRATE*const.JWHEEL*0.7);
+estimator.low_power= false;
+estimator.eclipse= env_eclipse(estimator.position_eci,estimator.sat2sun_eci);
+estimator.valid_gps=all(isfinite([estimator.position_eci; estimator.velocity_eci; estimator.time]));
+estimator.valid_target_gps=estimator.valid_gps && all(isfinite([estimator.target_position_eci; estimator.target_velocity_eci; estimator.time]));
+
+if estimator.valid_gps && ~estimator.eclipse && ~all(isfinite(estimator.sat2sun_body))
+    estimator.sat2sun_body= [0;0;-1]; % assume the sun is coming from the docking face
+    estimator.quat_body_eci=utl_triad(estimator.sat2sun_eci,estimator.mag_eci,estimator.sat2sun_body,estimator.magnetic_field_body);
+end
```
kylekrol: Throwing this in an attitude estimator function and gps estimation function may be a good idea?

MATLAB/update_FC_state.m
```diff
@@ -12,91 +12,149 @@
...
+    sat2target_r_body_hat= utl_rotateframe(estimator.quat_body_eci,sat2target_r_eci_hat);
+    r_hat_body= utl_rotateframe(estimator.quat_body_eci,r_hat_eci);
+    r_hat_cross_sat2sun_body=cross(estimator.sat2sun_body,r_hat_eci);
+    state.primary_current_direction_body= r_hat_body;
+    state.primary_desired_direction_body= [1;0;0];
+    state.secondary_current_direction_body= sat2target_r_body_hat;
+    state.secondary_desired_direction_body= [0;0;-1];
+end
+if strcmp(state.main_state,'????')
+    state.main_state='PROFIT';
```
kylekrol: What's `'PROFIT'`?

kylekrol (commented):
For the most part looks really good. I left a couple miscellaneous questions. I think my largest concern is some slight divergence from where I think the flight computer implementations of the controllers may have been going (particularly the pointer state's PD buffer).

I just want to make sure we're one the same page about the internals of the MATLAB controllers and CXX controllers so we don't have any issues integrating the CXX functions into MATLAB (which I plan on doing once the full suite of attitude controllers is complete).

Let me know what you think!

@stewartaslan for awareness.

MATLAB/recursive_least_squares.m
nhz2: This isn't being used right now. Its a test script I made to try out using recursive least squares to estimate the magnetometer bias vector.

MATLAB/sensor_reading.m
nhz2: Yeah, we have to calculate that ourselves.

MATLAB/sensor_reading.m
nhz2: I was just messing around with it.

kylekrol: I can try and dig up some old MATLAB scripts that I made to play around with it. From what I remember, the error was bounded much better with (about 5x better) with the threshold set to 4 and not 3. Although either way the standard div was `< 2` degrees or so.

MATLAB/sensor_reading.m
kylekrol: So `sensor_readings.target_velocity_ecef` will only live in MATLAB?

MATLAB/config.m
nhz2: I'm not sure, I wanted to have way to point even if `quat_body_eci` was NaN. Before, the derivative term was just the gyro, but I wanted to see what would happen if I just did finite differences.

nhz2: Though I guess this basically doing the same thing as estimating `w_body`

MATLAB/sensor_reading.m
nhz2: Below are some example pointing errors with 3 and 4

## 3
pointing_error_j_3

## 4
pointing_error_j_4

MATLAB/config.m
kylekrol: Ah okay. But in terms of the sorta version 1.0 of the GNC algorithms we're good keeping the finite difference estimation in the estimator and no buffer in the controller?

MATLAB/recursive_least_squares.m
kylekrol: Do you have a link to the source/algorithm somewhere that I could take a look at real quick? Just curious about it.

MATLAB/update_FC_state.m
```diff
@@ -12,91 +12,149 @@
...
+    r_hat_body= utl_rotateframe(estimator.quat_body_eci,r_hat_eci);
+    v_hat_body= utl_rotateframe(estimator.quat_body_eci,v_hat_eci);
+    r_hat_cross_sat2sun_body=cross(estimator.sat2sun_body,r_hat_eci);
+    state.primary_current_direction_body= r_hat_body;
+    state.primary_desired_direction_body= [1;0;0];
+    state.secondary_current_direction_body= r_hat_cross_sat2sun_body/norm(r_hat_cross_sat2sun_body);
+    state.secondary_desired_direction_body= [0;0;1];
+end
+if strcmp(state.main_state,'rendezvous')
+    if estimator.tumbling || estimator.low_power || estimator.eclipse || ~estimator.valid_gps || ~estimator.valid_target_gps
```
kylekrol: This may be temporary, but if everything is behaving nominally except for having target position, velocity would we want to go back to detumble or hold attitude with the full ADCS system? The latter would probably give us a better shot at getting comms, gps, etc

**EDIT**: Actually setting it to `'detumble'` will let it propagate through to `'get gps'`, `'get target orbit'`, etc on the next FC pass correct? If so, I think this is still good.

nhz2: Hopefully if cd gps lock is lost, the orbit estimator can just propagate the target orbit, that way it won't have to go into detumble.

nhz2:
To run the simulation in MATLAB, go to the MATLAB directory in psim, then run:

`run_tests` and make sure there are no failed tests.

`main` will generate the trajectories

`plot_pointing_errors(main_state_trajectory)` will show the pointing errors

`plot_state_transitions(computer_state_follower_trajectory)` will show the state of the follower flight computer.

there are other plotting functions in the `plot` folder

MATLAB/adcs/adcs_detumbler.m
```diff
@@ -0,0 +1,36 @@
...
+% Cornell University
+global const
+switch state.counter
+    case 5
+        %save magnetic field reading
+        state.old_magnetic_field= magnetometer_body;
+    case 9
+        %calculate finite difference and bang bang controller
+        deltaB= magnetometer_body-state.old_magnetic_field;
+        state.magrod_moment_cmd= -sign(deltaB)*const.MAXMOMENT;
```
tanishqaggarwal: What's the meaning of 5 and 9?

MATLAB/adcs/adcs_pointer.m
```diff
@@ -0,0 +1,100 @@
...
+% Space Systems Design Studio
+% Cornell University
+global const
+
+%% Get Error %%
+%get error, this is the first three terms of the error quaternion, if the
+%forth component is positive.
+error=zeros(3,1);
+if all(isfinite([secondary_current_direction_body; secondary_desired_direction_body])) ...
+        && abs(dot(secondary_current_direction_body,primary_current_direction_body)) < cos(10*pi/180) ...
```
tanishqaggarwal, line 62: Why is this norm constraint necessary?

MATLAB/adcs/adcs_pointer.m
```diff
@@ -0,0 +1,100 @@
...
+% end
+dterm= -const.ATTITUDE_PD_KD*rate_body;
+wheel_torque_cmd= -(pterm+dterm);
+
+%% Angular Momentum Control
+if (norm(angular_momentum_body)>0.05*const.MAXWHEELRATE*const.JWHEEL)
+    magrod_moment_cmd= sign(cross(angular_momentum_body,magnetic_field_body))*const.MAXMOMENT;
+else
+    magrod_moment_cmd= zeros(3,1);
+end
```
tanishqaggarwal, line 97: Does this code basically mean that the satellite is constantly reducing wheel momentum?

MATLAB/initialize_main_state.m
```diff
@@ -62,8 +61,8 @@
 actuators= actuators_off_state();
 
 % sensors
-sensors.gyro_bias= zeros(3,1);
-sensors.magnetometer_bias= zeros(3,1);
+sensors.gyro_bias= const.gyro_bias_sdiv*randn(3,1);
+sensors.magnetometer_bias= const.magnetometer_bias_sdiv*randn(3,1);
```
tanishqaggarwal, line 65: Is this randomnesss controlled?

MATLAB/sensor_reading.m
tanishqaggarwal: What is this j paramter?

MATLAB/utl/utl_ode2.m
```diff
@@ -34,17 +34,17 @@
...
+%   error(msg);  
+% end  
 
 y0 = y0(:);   % Make a column vector.
-if ~isequal(size(y0),size(f0))
-  error('Inconsistent sizes of Y0 and f(t0,y0).');
-end  
+% if ~isequal(size(y0),size(f0))
+%   error('Inconsistent sizes of Y0 and f(t0,y0).');
+% end
```
tanishqaggarwal: Why is this all commented?

MATLAB/update_FC_state.m
```diff
@@ -12,91 +12,196 @@
...
+    if true
+        state.on_time=uint64(0);
+        state.main_state='initialization hold';
+    end 
+end
+if strcmp(state.main_state,'initialization hold')
+    %move to next state if initialization time has passed
+    if state.initialization_hold_done || state.on_time>(30*60*1E9)
+        state.initialization_hold_done=true;%save this in eprom
+        state.main_state='detumble';
```
tanishqaggarwal, line 71: FYI Flight Software stores on_time instead of hold_done. The rationale is that the satellite can have low battery charge and can die during the post-deployment phase. I don't think it's necessary to copy this into sim though.

MATLAB/update_FC_state.m
```diff
@@ -12,91 +12,196 @@
...
+    end
+    %calculate pointing strategy
+    state.primary_current_direction_body= estimator.sat2sun_body/norm(estimator.sat2sun_body);
+    state.primary_desired_direction_body= [sqrt(2)/2;sqrt(2)/2;0];
+    state.secondary_current_direction_body= NaN(3,1);
+    state.secondary_desired_direction_body= NaN(3,1);
+end
+return2detumble= return2detumble || estimator.eclipse || ~estimator.valid_gps;
+for i=1:3
+    if strcmp(state.main_state,"calibrate magnetometer "+i)
```
tanishqaggarwal, line 100: When during the mission should we use a calibrate magnetometer state?

MATLAB/update_FC_state.m
```diff
@@ -12,91 +12,196 @@
...
+                state.main_state= 'get target orbit';
+            end
+        end 
+        %calculate pointing strategy
+        state.primary_current_direction_body= estimator.sat2sun_body/norm(estimator.sat2sun_body);
+        state.primary_desired_direction_body= axis; 
+        state.secondary_current_direction_body= NaN(3,1);
+        state.secondary_desired_direction_body= NaN(3,1);
+    end   
+end
```
tanishqaggarwal: Confused by the logic here in general. Could you explain what it's doing?

MATLAB/update_FC_state.m
```diff
@@ -12,91 +12,196 @@
...
+    end 
+    %calculate pointing strategy
+    r_hat_eci= estimator.position_eci/norm(estimator.position_eci);
+    r_hat_body= utl_rotateframe(estimator.quat_body_eci,r_hat_eci);
+    r_hat_cross_sat2sun_body=cross(estimator.sat2sun_body,r_hat_eci);
+    state.primary_current_direction_body= r_hat_body;
+    state.primary_desired_direction_body= [1;0;0];
+    if( norm(r_hat_cross_sat2sun_body) > sin(10*pi/180))
+        state.secondary_current_direction_body= r_hat_cross_sat2sun_body/norm(r_hat_cross_sat2sun_body);
+        state.secondary_desired_direction_body= [0;0;1];
```
tanishqaggarwal, line 137: For semantic reasons, might it be better for the secondary direction to be NaN?

MATLAB/update_FC_state.m
```diff
@@ -12,91 +12,196 @@
...
+    elseif estimator.valid_target_gps
+        state.main_state='rendezvous';
+    end 
+    %calculate pointing strategy
+    r_hat_eci= estimator.position_eci/norm(estimator.position_eci);
+    r_hat_body= utl_rotateframe(estimator.quat_body_eci,r_hat_eci);
+    r_hat_cross_sat2sun_body=cross(estimator.sat2sun_body,r_hat_eci);
+    state.primary_current_direction_body= r_hat_body;
+    state.primary_desired_direction_body= [1;0;0];
+    if( norm(r_hat_cross_sat2sun_body) > sin(10*pi/180))
```
tanishqaggarwal, line 135: I'll change the alignment threshold in Flight Software to also be 10 degrees

tanishqaggarwal: In general is there some way we could standardize these thresholds across Flight Software and sim? It would be great to not have to manually ensure their similitude

tanishqaggarwal (commented):
Looks great, I just have a few questions (mostly due to my ignorance of sim)

MATLAB/adcs/adcs_pointer.m
nhz2: The constraint says to ignore the the secondary pointing objective if it is within 10 degrees of being parallel or anti parallel to the primary, if its within 10 degrees, then just pointing the primary is all that can be done. This also prevents triad from failing.

MATLAB/adcs/adcs_pointer.m
nhz2: Yes, but its the full satellite angular momentum not just from the wheels.

MATLAB/initialize_main_state.m
nhz2: yes, in `initialize_main_state(seed,condition)` you can set the seed for the sim/ initial conditions, and the `run_sim(num_steps,sample_rate,main_state,computer_state_follower,computer_state_leader)` function outputs a `rng_state_trajectory` which can be used to reset the rng state to any point from the sim by doing `rng(rng_state_trajectory{40})` for example.

MATLAB/update_FC_state.m
nhz2: Right now I am calibrating the magnetometer after eclipse, and after detumbling.

MATLAB/update_FC_state.m
nhz2: The main idea is to rotate the sat around to estimate the magnetometer bias. There are three modes, each one points a different axis at the sun, so the mag bias on that axis can be estimated accurately.

MATLAB/update_FC_state.m
nhz2: Yeah, the secondary direction is NaN by default. Also, I haven't actually tested this mode, because the cdgps is perfect right now, so this mode is skipped. This can be tested more once the gps sensor model is better and there is an orbit estimator.

MATLAB/sensor_reading.m
nhz2: It's the minimum number of photodiodes above threshold to calculate a sun vector on the ADCS box.

MATLAB/update_FC_state.m
tanishqaggarwal: https://knowyourmeme.com/memes/profit

kylekrol (approved):
I like the organizational improvements! If this is good to go, merging it sooner rather than later would be cool so I can get on the rendezvous changes as well.

tanishqaggarwal (approved):
I'm going to approve bc my questions aren't that important and sounds like work is waiting on this PR

stewartaslan (commented):
Looks good to me.

stewartaslan (approved):

PR #93, opened 2019-12-07
Dev/ground software

fatimayousuf:
I made this document. Hopefully it helps: https://docs.google.com/document/d/1L3BIERQiuarFjygX8gYp65O_BY6mo1y6pgNc7L4zIkQ/edit


python/gsw/main.py
```diff
@@ -0,0 +1,303 @@
...
+        return json.loads(data)
+
+    def check_for_email(self):
+        '''
+        Checks the PAN email account. Updates the most recent MOMSN 
+        and MTMSN numbers. Returns the contents of the email attachement
+        of the most recent unread email from the Iridium email account.
+        '''
+        #look for all new emails from iridium
+        _, data = self.mail.search(None, '(FROM "fy56@cornell.edu")', '(UNSEEN)')
```
shihaocao: Should this be changed to something more general eventually? (already talked to Fatima about this)

python/usb_console/simulation.py
```diff
@@ -6,7 +6,7 @@
 import time, timeit
 import math
 import threading
-import matlab.engine
+#import matlab.engine
```
tanishqaggarwal: Can you uncomment this?

python/usb_console/radio_session.py
```diff
@@ -56,158 +61,62 @@ def __init__(self, device_name, datastore, logger, radio_keys_config):
         self.mtmsn=-1
```
tanishqaggarwal: I don't think we need the overriden variables; can you get rid of them?

python/gsw/main.py
```diff
@@ -0,0 +1,350 @@
+from flask import Flask, request, jsonify
+from flasgger import Swagger
+from flasgger.utils import swag_from
+from elasticsearch import Elasticsearch
```
tanishqaggarwal, line 4: Can you:
- Create a python venv for running Ground Software
- Add all these items to a requirements.txt file using pip? `pip freeze > requirements.txt`

It would be nice to also add a README so that everyone can set up ground software with a common set of instructions.

python/gsw/main.py
```diff
@@ -0,0 +1,350 @@
...
+        2) when we recieve an uplink confirmation.
+        - If there is a statefield report, index the statefield report and 
+        create and index an iridium report.
+        - If there is not a statefield report but we recently recieved an uplink
+        confirmation, then create and index an iridium report.
+        '''
+        # Connect to elasticsearch
+        es=Elasticsearch([{'host':self.es_server,'port':self.es_port}])
+
+        while self.run_email_thread==True:
```
tanishqaggarwal, line 199: This while loop will currently eat up a lot of CPU bandwidth because there's no waiting at the end of the loop. Consider adding a 1 second delay to the loop so that other computer processes can run in the meantime.

Can you also quantify how many resources this takes up? RAM, CPU, etc. This is important so we can scale down our Virtual Machine if possible

python/gsw/main.py
```diff
@@ -0,0 +1,350 @@
...
+    sf_res = es.index(index='statefield_report', doc_type='report', body=sf_report)
+    res={"Report Status": sf_res['result']}
+    return res
+
+# Endpoint for getting data from the statefield reports index in elasticsearch
+@app.route("/search-statefields", methods=["POST"])
+@swag_from("endpoint_configs/search_statefields_config.yml")
+def get_statefield():
+    # Connect to elasticsearch
+    es=Elasticsearch([{'host':readIr.es_server,'port':readIr.es_port}])
```
tanishqaggarwal: This object is declared several times over the course of the file. Is this a heavy object to construct? If it is, consider moving it to the top of the file so that it can be shared across all of the code.

python/gsw/main.py
```diff
@@ -0,0 +1,350 @@
...
+                    "order": "desc"
+                }
+            }
+        ],
+        "size": 1
+    }
+    res = es.search(index='statefield_report', body=json.dumps(search_object))
+    # Get the value of that statefield from the document
+    most_recent_field=res["hits"]["hits"][0]["_source"][field]
+    return most_recent_field
```
tanishqaggarwal: What's the behavior of this return statement if the field name doesn't exist?

python/gsw/main.py
```diff
@@ -0,0 +1,350 @@
...
+@app.route("/test", methods=["POST"])
+@swag_from("endpoint_configs/echo_config.yml")
+def echo():
+    input_json=request.get_json()
+    message=str(input_json["message"])
+    res={"Recieved": message}
+    return json.dumps(res)
+
+# Endpoint for indexing data in elasticsearch
+# Mostly for testing purposes. We don't use this to actually index data in elasticsearch
```
tanishqaggarwal, line 275: Does elasticsearch auto-index by default?

python/gsw/main.py
```diff
@@ -0,0 +1,350 @@
...
+swagger=Swagger(app, config=swagger_config)
+
+# Endpoint for testing post requests
+@app.route("/test", methods=["POST"])
+@swag_from("endpoint_configs/echo_config.yml")
+def echo():
+    input_json=request.get_json()
+    message=str(input_json["message"])
+    res={"Recieved": message}
+    return json.dumps(res)
```
tanishqaggarwal: I think we can remove this, now that you've tested this function

python/gsw/main.py
```diff
@@ -0,0 +1,350 @@
...
+import base64
+import os
+import email
+import logging
+
+app = Flask(__name__)
+app.config["SWAGGER"]={"title": "PAN Ground Software", "uiversion": 2}
+ 
+class read_iridium(object):
+    def __init__(self, radio_keys_config, server_keys_config):
```
tanishqaggarwal: One major feature we absolutely need is separating telemetry by IMEI number. Please consider doing this soon (this is higher priority than the EEPROM controller.)

python/usb_console/radio_session.py
tanishqaggarwal: I think we should fetch the IMEI number as a separate constructor argument, rather than from radio_keys_config. The argument should be populated with the value from standard config data, e.g. see `fc_only_native.json` for an example. The reason for this is that the IMEI number is declared in the standard config files in order to tie an IMEI number to the name of a Flight Device.

python/usb_console/run_simulation.py
```diff
@@ -210,6 +208,8 @@ def stop_all(self, reason_for_stop):
             radios_config = config_data["radios"]
         with open(os.path.join(os.path.dirname(os.path.abspath(__file__)), 'configs/radio_keys.json')) as radio_keys_config_file:
             radio_keys_config = json.load(radio_keys_config_file)
+        with open(os.path.join(os.path.dirname(os.path.abspath(__file__)), 'configs/flask_keys.json')) as flask_keys_config_file:
+            flask_keys_config = json.load(flask_keys_config_file)
```
tanishqaggarwal: It would be nice if these config paths were read in as command-line arguments, like the base config file is read. Not necessary though.

python/usb_console/radio_session.py
```diff
@@ -10,6 +10,7 @@
 import base64
 import os
 import email
+import requests
```
tanishqaggarwal, line 9: Remember to add to requirements.txt as part of the venv setup (check out the other comment I've made about this on gsw/main.py)

tanishqaggarwal (changes requested):
Looks really, really good. Made a couple of suggestions.

python/gsw/main.py
fatimayousuf: Not entirely sure what you mean by auto-indexing. I index information in the post_to_es() function, every time I call self.es.index(). If I try to index a document in an index that doesn't exist, then elasticsearch will automatically create the index.

python/gsw/main.py
```diff
@@ -81,8 +89,6 @@ def check_for_email(self):
             #.fetch() fetches the mail for given id where 'RFC822' is an Internet 
             # Message Access Protocol.
             _, data = self.mail.fetch(num,'(RFC822)')
-            #mark the email message as read.
-            self.mail.store(num, '+FLAGS', '(\\Seen)')
```
tanishqaggarwal: Why did you stop marking the messages as read?

python/gsw/main.py
```diff
@@ -93,6 +99,9 @@ def check_for_email(self):
 
                     #handles uplink confirmations
                     if email_subject.find("SBD Mobile Terminated Message Queued for Unit: ")==0:
```
tanishqaggarwal: It would be nice if these two email processings could be unit-tested. We can defer this though

python/gsw/main.py
tanishqaggarwal: Why did you stop marking emails as read?

python/gsw/main.py
tanishqaggarwal, line 94: Can you add unit testing for this email text processing stuff?

python/usb_console/radio_session.py
```diff
@@ -45,169 +51,53 @@ def __init__(self, device_name, datastore, logger, radio_keys_config):
...
-        
-        Read the latest value of the state field received via downlink with the name "field".
+        Read state by posting a request for data to the Flask server
         '''
-        return self.statefields[field]
+        headers = {
+            'Accept': 'text/html',
+        }
+
+        response = requests.get('http://'+self.flask_server+':'+self.flask_port+'/search-statefields?imei='+str(self.imei)+'&field='+str(field), headers=headers)
```
tanishqaggarwal: This kind of string building will fail if the inputs are not correctly URL-encoded. For example, if `field` contains a period, the request URL will be malformed.

There is a cleaner and safer way to do this within `requests`. See "Making a POST request" on https://www.geeksforgeeks.org/get-post-requests-using-python/

python/usb_console/radio_session.py
```diff
@@ -45,169 +51,53 @@ def __init__(self, device_name, datastore, logger, radio_keys_config):
...
-        #create a JSON file with the updated statefields and send it to the iridium email
-        with open('uplink.json', 'w') as json_uplink:
-            json.dump(updated_fields, json_uplink)
-        if self.send_uplinks==True:
-            yag.send('sbdservice@sbd.iridium.com', 'Data', 'uplink.json')
+        headers = {
+            'Accept': 'text/html',
+        }
+
+        response = requests.get('http://'+self.flask_server+':'+self.flask_port+'/search-iridium?imei='+str(self.imei)+'&field=send-uplinks', headers=headers)
```
tanishqaggarwal: Same comment about string-building

python/usb_console/radio_session.py
```diff
@@ -10,6 +10,7 @@
 import base64
 import os
 import email
```
tanishqaggarwal: We can probably pare down a good number of these import statements, now that a lot of radio session logic got moved to ground software

python/gsw/main.py
```diff
@@ -0,0 +1,370 @@
...
+import os
+import email
+import time
+import logging
+
+app = Flask(__name__)
+app.config["SWAGGER"]={"title": "PAN Ground Software", "uiversion": 2}
+ 
+class read_iridium(object):
+    def __init__(self, radio_keys_config, server_keys_config):
```
tanishqaggarwal: The number of configs have grown a lot in the past couple of weeks. Can you add a README to the config folder explaining the purpose of the config files?

python/gsw/main.py
```diff
@@ -0,0 +1,370 @@
...
+                    "confirmation-mtmsn": self.confirmation_mtmsn,
+                    "send-uplinks": self.send_uplinks,
+                    "time": str(datetime.now().isoformat())
+                })
+
+                # Index iridium report in elasticsearch
+                iridium_res = self.es.index(index='iridium_report_'+str(self.imei), doc_type='report', body=ir_report)
+                # Print whether or not indexing was successful
+                print("Iridium Report Status: "+iridium_res['result']+"\n\n")
+                # Record that we have not recently recieved any uplinks as we just indexed the most recent one
```
tanishqaggarwal: There is duplicated Iridium report code here. I think it would be better to restructure like this:

- Add Iridium report, always
- Add state field report if it exists
- If uplink confirmation is received, set the flag to true

python/gsw/main.py
```diff
@@ -0,0 +1,370 @@
...
+    "swagger_ui":True,
+    "specs_route":"/swagger/"
+}
+swagger=Swagger(app, config=swagger_config)
+
+# Endpoint for indexing data in elasticsearch
+# Mostly for testing purposes. We don't use this to actually index data in elasticsearch
+@app.route("/telemetry", methods=["POST"])
+@swag_from("endpoint_configs/telemetry_config.yml")
+def index_sf_report():
```
tanishqaggarwal, line 278: It's not possible to add to a particular radio's data because this endpoint doesn't take IMEI number into account. This impeded testing for me. Can you add that in?

python/gsw/main.py
```diff
@@ -0,0 +1,370 @@
...
+import json
+import imaplib
+import base64
+import os
+import email
+import time
+import logging
+
+app = Flask(__name__)
+app.config["SWAGGER"]={"title": "PAN Ground Software", "uiversion": 2}
```
tanishqaggarwal: Can you move these two lines down closer to the `if __name__ == "__main__"` line? Preferably, it should all be within the if-block. Right now you're alternating between defining classes/methods and instantiating them; it's easier to read if we define all the stuff first, then instantiate it.

python/gsw/main.py
```diff
@@ -0,0 +1,370 @@
...
+            "route":"/apispec_1.json",
+            "rule_filter":lambda rule:True,
+            "model_filter":lambda tag:True
+        }
+    ],
+    "static_url_path": "/flassger_static",
+    "swagger_ui":True,
+    "specs_route":"/swagger/"
+}
+swagger=Swagger(app, config=swagger_config)
```
tanishqaggarwal: Same comment as above about moving instantiation closer to `__name__ == "__main__"`

python/gsw/main.py
```diff
@@ -0,0 +1,370 @@
...
+        '''
+        Reads the most recent unread email from the Iridium email account. 
+
+        If it is a downlink, the function returns a statefield report. If it is an uplink 
+        confirmation, the function records that we recieved an uplink confirmation and returns
+        None. If there is neither a downlink or an uplink, the function returns None.
+
+        If either a downlink or an uplink is recieved, the function will update the most recent MOMSN 
+        and MTMSN numbers and check whether or not radioSession can send uplinks.
+        '''
```
tanishqaggarwal, line 75: Great job with this documentation. I'm loving the detailed commenting.

python/gsw/main.py
```diff
@@ -0,0 +1,370 @@
...
+                                    if line.find("MTMSN")!=-1:
+                                        self.mtmsn=int(line[line.find("MTMSN")+9:line.find("MTMSN")+11])
+
+                            #Verify whether or not RadioSession can send uplinks
+                            if self.confirmation_mtmsn != self.mtmsn:
+                                #stop radio session from sending any more uplinks
+                                self.send_uplinks=False
+                            else:
+                                #allow radio session to send more uplinks
+                                self.send_uplinks=True
```
tanishqaggarwal, line 119: I don't think we should do this here. Here's the problem I see: if `confirmation_mtmsn == self.mtmsn`, then it means that Iridium has queued up the most recent uplink, but it doesn't mean that the _satellite_ has received the uplink. Setting `send_uplinks` to true will therefore allow uplink packets to stack up on the Iridium servers, since RadioSession will just see a go-ahead to send more packets.

Maybe my logic is wrong here, though. Let me know if this sounds wrong

python/gsw/main.py
```diff
@@ -0,0 +1,370 @@
...
+                            # Get the body text of the email and look for the MOMSN/MTMSN number
+                            if part.get_content_type() == "text/plain":
+                                email_body = part.get_payload(decode=True).decode('utf8')
+                                for line in email_body.splitlines():
+                                    if line.find("MOMSN")!=-1:
+                                        self.momsn=int(line[7:])
+                                    if line.find("MTMSN")!=-1:
+                                        self.confirmation_mtmsn=int(line[7:])
+                            
+                            #Verify whether or not radioSession can send uplinks
```
tanishqaggarwal: Should be "Set whether or not"

python/gsw/main.py
```diff
@@ -0,0 +1,370 @@
...
+
+                            if part.get('Content-Disposition') is None:
+                                continue
+
+                            #get the body text of the email and look for the MOMSN/MTMSN number
+                            if part.get_content_type() == "text/plain":
+                                email_body = part.get_payload(decode=True).decode('utf8')
+                                for line in email_body.splitlines():
+                                    if line.find("MTMSN")!=-1:
+                                        self.mtmsn=int(line[line.find("MTMSN")+9:line.find("MTMSN")+11])
```
tanishqaggarwal, line 111: You use the exact same code below in processing downlinks. Consider breaking it out to a function, so that we minimize code duplication. (Breaking it out into a function will also make things easier to read)

python/gsw/main.py
tanishqaggarwal: Marking as "unresolved" because I'm interested in these numbers

tanishqaggarwal: If the CPU numbers are high (I suspect they are, given how much your fan spins up during work sessions...:P), I honestly think we can let the delay be as large as 10 seconds. I think 10 seconds is a sweetspot because we won't see a noticeable reduction in resource utilization by increasing beyond 10 seconds, and emails are only going to come once every few hours, so running this super frequently isn't a good idea

python/usb_console/radio_session.py
```diff
@@ -45,169 +51,53 @@ def __init__(self, device_name, datastore, logger, radio_keys_config):
...
-        with open('uplink.json', 'w') as json_uplink:
-            json.dump(updated_fields, json_uplink)
-        if self.send_uplinks==True:
-            yag.send('sbdservice@sbd.iridium.com', 'Data', 'uplink.json')
+        headers = {
+            'Accept': 'text/html',
+        }
+
+        response = requests.get('http://'+self.flask_server+':'+self.flask_port+'/search-iridium?imei='+str(self.imei)+'&field=send-uplinks', headers=headers)
+        print(response.text)
```
tanishqaggarwal: Can you delete this print statement?

tanishqaggarwal: Actually, this information is useful for debugging. Can you instead add it to `self.logger`

python/gsw/main.py
tanishqaggarwal: I noticed below that you're using `readIr.es` as a reference to an ElasticSearch object within your HTTP handlers. To get around this issue, you might be able to add `es` as an attribute of `app` instead (I think this is also inherently cleaner)

python/gsw/main.py
```diff
@@ -0,0 +1,370 @@
...
+    # Get the value of that statefield from the document
+    if es.indices.exists(index='iridium_report_'+str(imei)):
+        res = es.search(index='iridium_report_'+str(imei), body=json.dumps(search_object))
+        if len(res["hits"]["hits"])!=0:
+            most_recent_field=res["hits"]["hits"][0]["_source"][field]
+            return str(most_recent_field)
+        else:
+            return "Unable to find field in index: "+"iridium_report_"+str(imei)
+    else:
+        return "Unable to find index: "+"iridium_report_"+str(imei)
```
tanishqaggarwal: Consider using "f" notation for string building https://realpython.com/python-f-strings/

python/gsw/main.py
```diff
@@ -0,0 +1,370 @@
...
+    # Get the value of that statefield from the document
+    if es.indices.exists(index='statefield_report_'+str(imei)):
+        res = es.search(index='statefield_report_'+str(imei), body=json.dumps(search_object))
+        if len(res["hits"]["hits"])!=0:
+            most_recent_field=res["hits"]["hits"][0]["_source"][statefield]
+            return most_recent_field
+        else:
+            return "Unable to find field in index: "+"statefield_report_"+str(imei)
+    else:
+        return "Unable to find index: "+"statefield_report_"+str(imei)
```
tanishqaggarwal: You're using very similar code below. Consider breaking this out into a function

tanishqaggarwal (changes requested):
I made some more suggestions.

python/gsw/endpoint_configs/telemetry_config.yml
```diff
@@ -0,0 +1,28 @@
...
+          example: "11/12/13"
+        field:
+          type: "string"
+          example: "control cycle count"
+        value:
+          type: "integer"
+          example: 6
+responses: 
+  405:
+    description: "Invalid input"
```
tanishqaggarwal: When do you return 405 if the input is invalid? I couldn't find it in the code, I probably just missed it though

tanishqaggarwal:
Feel free to update the root-level README to indicate that your Ground Software is now a part of this repository  :)

python/gsw/main.py
fatimayousuf: It didn't make a difference. Once the function reads the email message, the "read" flag in the email message is automatically set to True anyways.

tanishqaggarwal: Gotcha.

python/gsw/main.py
fatimayousuf: Do you mean we should be adding an Iridium report every 10 seconds? I still think it makes more sense to make iridium reports whenever we receive either a downlink or an uplink confirmation, since those are the only situations in which the momsn and mtmsn numbers would be changed.

python/gsw/main.py
fatimayousuf: I can move it down to just before I declare the endpoints and the SwaggerUI API, but not after, otherwise I'll get an error.

python/gsw/main.py
fatimayousuf: So this is my understanding of how the Iridium email system works (I could be wrong, though): 

The uplink confirmation email gives the mtmsn of the message queued to be be sent to the satellite by Iridium. If the function identifies an uplink confirmation email, it will set the mtmsn class variable to the mtmsn of the most recent message queued for uplink.

The downlink email gives the mtmsn of most recent message that the satellite recieved. If the function identifies a downlink, it sets the confirmation_mtmsn variable to the mtmsn of the most recent message that the satellite recieved.

If the mtmsn and the confirmation_mtmsn match up, then radio session is cleared to send uplinks. I think that, by this system, we can be sure that we are only queuing one message to be sent to the satellite at a time.

python/gsw/main.py
tanishqaggarwal: Oops!  I misread the code, and you're right, I think that does make more sense. I still think we should break out the redundant code into a function, though.

python/gsw/main.py
tanishqaggarwal: That makes sense.

python/gsw/main.py
tanishqaggarwal: This makes sense! Sorry, I got confused between the meanings of `confirmation_mtmsn` and `mtmsn`.

Can the code that checks these variables' equality be moved below the email processing code? Otherwise it's duplicated across the uplink and downlink.

python/gsw/main.py
tanishqaggarwal: Nevermind, the code isn't equal! I misread.

python/gsw/main.py
fatimayousuf: I can't check the variables equality after I return the statefield report. Since this function only looks at the most recent email, we will be checking equality only once each time this function is called. So I don't think having it duplicated should be much of a problem.

python/gsw/main.py
fatimayousuf: If I broke all this out into a separate function, I would basically be creating a single endpoint to handle searching both iridium and statefield reports in elasticsearch. If I combined these two endpoints into a single endpoint, I would have to add another query parameter specifying whether to search the iridium index or the statefield index. Personally I think it is simpler to keep two separate endpoints for searching statefield reports and iridium reports and avoid adding that extra parameter. Do you think it would be better to combine them into a single endpoint?

fatimayousuf: Upon further reflection I realized i could just combine these two into a single endpoint

python/gsw/endpoint_configs/telemetry_config.yml
fatimayousuf: The endpoint already returns a different error message when an index or field can't be found, so yeah this doesn't really do anything. I'll get rid of it

python/usb_console/configs/README.md
```diff
@@ -20,3 +20,34 @@ Here you can find some template configurations. I recommend copy-pasting a templ
   - `connect` If false, no attempt is made to establish the radio connection. This option exists so that the simulation can be run separately from radio testing.
 
 All of the above fields are required, but the `devices` and `radios` list may potentially be empty.
+
```
tanishqaggarwal, line 23: Nice! Thanks

tanishqaggarwal (changes requested):
Seems like you forgot to update telemetry_config.yml

tanishqaggarwal (dismissed):

tanishqaggarwal (approved):

athena255 (approved):

Issue #34, opened 2019-10-17
Rendezvous Algorithm


kylekrol:
Eventually, everything that has to do with rendezvous.

tanishqaggarwal:
Important: we need to discretize the rendezvous algorithm so that we can run sections of it in every control cycle. Sections should take at most ~20 ms to run.

kylekrol:
@tanishqaggarwal I would hope that the algorithm will run faster than that and we wouldn't need to space it out. That being said, in it's current state it'll be easy to separate out.

kylekrol:
Closing in favor of newer orbit controller issues.

PR #94, opened 2019-12-09
Add Initial Rendezvous Algorithm Implementation in MATLAB with Simulation

kylekrol:
Relates to #34.

Opening this up to keep master up to date with rendezvous code. This will get cleaned up in the next big PR where I'll hopefully have some up code to run multiple simulations in parallel to get some Monte Carlo results (a single rendezvous sim already takes ~5/10 minutes to complete).

I will also be breaking out CW and ECI <-> Hill/LVLH conversions into utility functions in the future!


nhz2 (approved):
Looks good. It's amazing that it works in J4 with the clohessy wiltshire approximation.

PR #96, opened 2019-12-14
Adding USB console features

tanishqaggarwal:
- Automatic logger saving to file
- Datastore is now based on TinyDB rather than a plain JSON file (enables file caching, which is great for memory use)
- Config validation
- Move datastore/logger construction to within StateSesssions, rather than RunSimulation
- **Adds plotting** to the USB console
- Plotting data after the simulation has ended is possible

**Note that the invocation for running the simulation has changed.**

This is a precursor to the PR with psim-FSW ADCS estimator integration.


python/usb_console/plotter.py
```diff
@@ -0,0 +1,163 @@
...
+            if field_data[0][1] in ["true", "false"]:
+                # It's a boolean
+                data_vals = [(1 if datapoint == "true" else 0)
+                             for datapoint in field_data]
+            else:
+                try:
+                    # It might be an integer
+                    data_vals = [int(datapoint[1]) for datapoint in field_data]
+                except ValueError:
+                    try:
```
kylekrol, line 89: Not a huge deal atm and by no means should it stop the PR in my opinion, but we may want a cleaner way of representing type in the future. For example, this wouldn't really allows us to represent a `vec2` it seems.

kylekrol (approved):

PR #107, opened 2020-01-20
Added Rendezvous Related Utility Functions

kylekrol:
The functions were tested with the `MATLAB/rendezvous/rendezvous.m` script.


MATLAB/utl/utl_orbrate.m
```diff
@@ -0,0 +1,8 @@
+function n = utl_orbrate(a)
+% Calculates the orbital rate of the sattelite assuming a circular orbit
```
nhz2: Add what units a and n are to the description.

nhz2 (changes requested):
Looks good, just one issue with units in the docs. Also maybe some of these functions should be in the rendezvous folder, or a helper function sub folder, if they are only used for rendezvous, like the l1cost functions.

MATLAB/utl/utl_orbrate.m
kylekrol: Yeah good point - MKS for everything.

kylekrol:
I'm gonna migrate some of the rendezvous specific code to `MATLAB/rendezvous` as suggested and will fix that comment too.

nhz2 (approved):

PR #119, opened 2020-02-08
Gomspacecheckout

afoarce:
Hardware checkout case for the Gomspace battery that reads its state fields and compares to known working ranges. Outputs are written to the simulation log and errors are printed to the console.


python/ptest/cases/gomspace_checkout_case.py
```diff
@@ -0,0 +1,104 @@
...
+        # readable fields
+
+        vboost = [
+            simulation.flight_controller.read_state("gomspace.vboost.output1"),
+            simulation.flight_controller.read_state("gomspace.vboost.output2"),
+            simulation.flight_controller.read_state("gomspace.vboost.output3")]
+
+        vbatt = simulation.flight_controller.read_state("gomspace.vbatt")
+        if vbatt < 6.0 or vbatt > 8.4:
+            print("Vbatt is out of expected range [6, 8.4] at: " + vbatt)
```
tanishqaggarwal: vbatt is in units of mV.

python/ptest/cases/gomspace_checkout_case.py
```diff
@@ -0,0 +1,104 @@
...
+
+        currin = [
+            simulation.flight_controller.read_state("gomspace.curin.output1"),
+            simulation.flight_controller.read_state("gomspace.curin.output2"),
+            simulation.flight_controller.read_state("gomspace.curin.output3")]
```
tanishqaggarwal: You can probably write a list comprehension on the state field name to shorten the code here? Same with the other repetitive state fields below.

python/ptest/cases/gomspace_checkout_case.py
```diff
@@ -0,0 +1,104 @@
...
+            simulation.flight_controller.read_state("gomspace.output.output3"),
+            simulation.flight_controller.read_state("gomspace.output.output4"),
+            simulation.flight_controller.read_state("gomspace.output.output5"),
+            simulation.flight_controller.read_state("gomspace.output.output6")]
+
+        for n in range(0, len(output)):
+            out_n = output[n]
+            if out_n < 0.5 or out_n > 3.0:
+                print("Output-" + n +
+                      " is out of range [0.3, 0.5] at: " + out_n)
```
tanishqaggarwal: The outputs have a value of either 0 or 1, as far as I know. I think a better test would be to ensure that the expected outputs are on (you'd have to dig into umbilical schematics to see which outputs these are supposed to be.)

python/ptest/cases/gomspace_checkout_case.py
```diff
@@ -0,0 +1,104 @@
...
+
+        wdt_i2c_time_left = simulation.flight_controller.read_state(
+            "gomspace.wdt_i2c_time_left")
+        counter_wdt_i2c = simulation.flight_controller.read_state(
+            "gomspace.counter_wdt_i2c")
+        counter_boot = simulation.flight_controller.read_state(
+            "gomspace.counter_boot")
```
tanishqaggarwal: We want to check that the time left is greater than 99 (the WDT timer is currently configured for 100 seconds)

We can't really write tests for the other two values, maybe just print them out to the screen?

python/ptest/cases/gomspace_checkout_case.py
```diff
@@ -0,0 +1,104 @@
...
+
+        temp = [
+            simulation.flight_controller.read_state("gomspace.temp.output1"),
+            simulation.flight_controller.read_state("gomspace.temp.output2"),
+            simulation.flight_controller.read_state("gomspace.temp.output3"),
+            simulation.flight_controller.read_state("gomspace.temp.output4")]
```
tanishqaggarwal: Maybe check these are near room temperature?

python/ptest/cases/gomspace_checkout_case.py
```diff
@@ -0,0 +1,104 @@
...
+
+        bootcause = simulation.flight_controller.read_state(
+            "gomspace.bootcause")
+        battmode = simulation.flight_controller.read_state("gomspace.battmode")
+        pptmode = simulation.flight_controller.read_state("gomspace.pptmode")
```
tanishqaggarwal: These aren't really testable values, we can print them maybe?

python/ptest/cases/gomspace_checkout_case.py
```diff
@@ -0,0 +1,104 @@
...
+        simulation.flight_controller.read_state(
+            "gomspace.power_cycle_output2_cmd")
+        simulation.flight_controller.read_state(
+            "gomspace.power_cycle_output3_cmd")
+        simulation.flight_controller.read_state(
+            "gomspace.power_cycle_output4_cmd")
+        simulation.flight_controller.read_state(
+            "gomspace.power_cycle_output5_cmd")
+        simulation.flight_controller.read_state(
+            "gomspace.power_cycle_output6_cmd")
```
tanishqaggarwal: We want to test that power cycling actually _works_. We should command these values and see if the output values change as we would expect them to.

python/ptest/cases/gomspace_checkout_case.py
```diff
@@ -0,0 +1,104 @@
...
+
+        simulation.flight_controller.read_state("gomspace.pptmode_cmd")
+
+        simulation.flight_controller.read_state("gomspace.heater_cmd")
+
+        simulation.flight_controller.read_state("gomspace.counter_reset_cmd")
+
+        simulation.flight_controller.read_state("gomspace.gs_reset_cmd")
+
+        simulation.flight_controller.read_state("gomspace.gs_reboot_cmd")
```
tanishqaggarwal: Let's test that these "command" functions actually work, rather than just reading their state.

python/ptest/cases/gomspace_checkout_case.py
```diff
@@ -0,0 +1,104 @@
...
+        simulation.flight_controller.read_state(
+            "gomspace.power_cycle_output4_cmd")
+        simulation.flight_controller.read_state(
+            "gomspace.power_cycle_output5_cmd")
+        simulation.flight_controller.read_state(
+            "gomspace.power_cycle_output6_cmd")
+
+        simulation.flight_controller.read_state("gomspace.pv1_cmd")
+        simulation.flight_controller.read_state("gomspace.pv2_cmd")
+        simulation.flight_controller.read_state("gomspace.pv3_cmd")
```
tanishqaggarwal: We don't really expect to use these particular commands in orbit so we might as well just not test these.

tanishqaggarwal (changes requested):
I realized I severely underspecified the nature of the test case. I made some suggestions on what stuff it could additionally test.

Before merging the suggestions, I think we should test on a real Gomspace! Let's save the console messages from the test so that we have hard verification evidence of having run this test.

python/ptest/cases/gomspace_checkout_case.py
```diff
@@ -15,90 +15,102 @@ def run_case_singlesat(self, simulation):
...
+
+        heater_cmd = read_state(self, "gomspace.heater_cmd")
+        heater_cmd_updated = write_state(
+            self, "gomspace.heater_cmd", not heater_cmd)
+        if heater_cmd == heater_cmd_updated:
+            print("Could not update heater")
+
+        counter_reset_cmd = read_state(self, "gomspace.counter_reset_cmd")
+        counter_reset_cmd_updated = write_state(
+            self, "gomspace_reset_cmd", not counter_reset_cmd)
```
tanishqaggarwal: Miscopied the reset cmd

python/ptest/cases/gomspace_checkout_case.py
```diff
@@ -15,90 +15,102 @@ def run_case_singlesat(self, simulation):
...
         # readable fields
 
-        vboost = [
-            simulation.flight_controller.read_state("gomspace.vboost.output1"),
-            simulation.flight_controller.read_state("gomspace.vboost.output2"),
-            simulation.flight_controller.read_state("gomspace.vboost.output3")]
+        vboost = [read_state(self, "gomspace.vboost.output" + str(i))
+                  for i in range(1, 4)]
 
         vbatt = simulation.flight_controller.read_state("gomspace.vbatt")
```
tanishqaggarwal: You probably need to cast vbatt to int.

python/ptest/cases/gomspace_checkout_case.py
```diff
@@ -15,90 +15,102 @@ def run_case_singlesat(self, simulation):
         simulation.cycle_no = simulation.flight_controller.read_state(
```
tanishqaggarwal, line 20: Can we add test case _failures_ as exception throws?

python/ptest/cases/gomspace_checkout_case.py
```diff
@@ -15,90 +15,102 @@ def run_case_singlesat(self, simulation):
...
+            # TODO: check umbilical for which outputs should be 0 and 1
             if out_n < 0.5 or out_n > 3.0:
                 print("Output-" + n +
                       " is out of range [0.3, 0.5] at: " + out_n)
 
         wdt_i2c_time_left = simulation.flight_controller.read_state(
             "gomspace.wdt_i2c_time_left")
+        if wdt_i2c_time_left > 99:
+            print("wdt_i2c_time_left is greater than 99 seconds at: " +
+                  str(wdt_i2c_time_left))
```
tanishqaggarwal: We do want the time to be greater than 99. We don't want it to be less than 99.

python/ptest/cases/gomspace_checkout_case.py
```diff
@@ -15,90 +15,102 @@ def run_case_singlesat(self, simulation):
...
-        pptmode = simulation.flight_controller.read_state("gomspace.pptmode")
+        print("counter_wdt_i2c is: " + str(counter_boot))
+
+        temp = [read_state(self, "gomspace.temp.output" + str(i))
+                for i in range(1, 5)]
+        for n in range(0, len(temp)):
+            temp_n = temp[n]
+            if temp_n < 20 or out_n > 22:
+                print("Temp-" + n +
+                      " is out of room temperature range [20, 22] degC at: " + str(temp_n))
```
tanishqaggarwal: Temp ends up being frequently 23-24 degrees in the lab. I think 20-25 is a safe bet.

tanishqaggarwal (changes requested):

python/ptest/cases/gomspace_checkout_case.py
```diff
@@ -0,0 +1,126 @@
...
+    def run_case_singlesat(self, simulation):
+        simulation.cycle_no = simulation.flight_controller.read_state(
+            "pan.cycle_no")
+
+        def read_state(self, string_state):
+            return simulation.flight_controller.read_state(string_state)
+
+        def write_state(self, string_state, state_value):
+            simulation.flight_controller.write_state(string_state, state_value)
+            return read_state(self, string_state)
```
fatimayousuf, line 28: Aren't you supposed to use self.read_state(string_state) to call a class method from within a class? idk maybe this works too lol

python/ptest/cases/gomspace_checkout_case.py
```diff
@@ -0,0 +1,126 @@
...
+        output = [bool(read_state(self, "gomspace.output.output" + str(i)))
+                  for i in range(1, 7)]
+
+        for n in range(0, len(output)):
+            out_n = output[n]
+            # checked umbilical for which outputs should be 0 and 1:
+            # 1-5 are all 5V, 6 is 3.3V
+            if out_n is False:
+                print("Output-" + n +
+                      " is not on: ")
```
fatimayousuf: Judging by the colon here, I feel like you _might_ have wanted to add something more to the print statement and maybe forgot. It doesn't matter to me either way, but as a good citizen with a civic responsibility to my fellow (wo)man, I feel that I must bring this matter to your attention.

fatimayousuf (approved):

python/ptest/cases/gomspace_checkout_case.py
```diff
@@ -0,0 +1,126 @@
...
+        counter_wdt_i2c = read_state(self, "gomspace.counter_wdt_i2c")
+        print("counter_wdt_i2c is: " + str(counter_wdt_i2c))
+        counter_boot = read_state(self, "gomspace.counter_boot")
+        print("counter_wdt_i2c is: " + str(counter_boot))
+
+        temp = [int(read_state(self, "gomspace.temp.output" + str(i)))
+                for i in range(1, 5)]
+        for n in range(0, len(temp)):
+            temp_n = temp[n]
+            if temp_n < 20 or out_n > 22:
```
fatimayousuf: I think you need to change this to n>25, based on the print statement below this

python/ptest/cases/gomspace_checkout_case.py
```diff
@@ -0,0 +1,126 @@
...
+
+        bootcause = read_state(self, "gomspace.bootcause")
+        print("bootcause is: " + str(bootcause))
+        battmode = read_state(self, "gomspace.battmode")
+        print("battmode is: " + str(battmode))
+        pptmode = read_state(self, "gomspace.pptmode")
+        print("pptmode is: " + str(pptmode))
+
+        # writable fields
+        power_cycle_output_cmd = [read_state(self, "gomspace.power_cycle_output" + str(i) + "_cmd")
```
fatimayousuf: By default the power cycle commands should be false. You can set each of them to true to start a power cycle for a single output. Once the power cycle is over, the command statefields should automatically be false again

fatimayousuf (commented):
Aaaah I'm sorry I clicked approve instead of request changes

python/ptest/cases/gomspace_checkout_case.py
```diff
@@ -0,0 +1,126 @@
...
+                                  for i in range(1, 7)]
+        power_cycle_output_cmd_updated = [write_state(self, "gomspace.power_cycle_output" + str(i) + "_cmd",
+                                                      not power_cycle_output_cmd[i])
+                                          for i in range(0, len(power_cycle_output_cmd))]
+        for n in range(0, len(power_cycle_output_cmd)):
```
fatimayousuf, line 126: It will take many control cycles (around 30-40 seconds or 300-400 control cycles) for a power cycle for a single output to complete. Maybe you can use two while loops to wait for outputs to turn off and then on again and have a maximum wait of 600 control cycles (1 min)?

fatimayousuf (commented):
Other than the power cycle test, the checkout case seems fine to me!

python/ptest/cases/gomspace_checkout_case.py
```diff
@@ -0,0 +1,153 @@
...
+        pptmode = read_state(self, "gomspace.pptmode")
+        print("pptmode is: " + str(pptmode))
+
+        # writable fields
+        power_cycle_output_cmd = [self.str_to_bool(read_state(self, "gomspace.power_cycle_output" + str(i) + "_cmd"))
+                                  for i in range(1, 7)]
+        cycle_no_init = int(read_state(self, "pan.cycle_no"))
+        cycle_no = cycle_no_init
+        print(power_cycle_output_cmd)
+        # wait for power cycling to be off
```
fatimayousuf: The power cycle command statefield is a command that will turn off and then turn on the gomspace outputs. When the gomspace control task is initialized, all the power cycle output commands are all set to false. So the first while loop here is kind of unnecessary because the power cycle commands should all be initialized to false anyway. 

You can start a power cycle for a specific output by setting the power cycle command statefield for that output to true. If a power cycle command is true, that means that the gomspace is currently power cycling that output, but it _doesn't_ tell you whether or not that output is currently on or off. You can check whether or not the output is currently on or off by reading the gomspace.output.outputX statefields. 

So you would have to set all the power cycle output commands to true to start a power cycle. And _then_ you would wait for the outputs to turn off within 30-60 seconds by reading the gomspace.output.outputX statefields. 

Once they are all off, then the gomspace control task should then start to turn the outputs currently power cycling on again. Then this is where you again would wait for all the outputs turn on. The outputs turn on fairly quickly. I can't imagine it taking more than 5 seconds.

After the output has finished turning off and then on again, the control task will set the power cycle command statefield to false.

python/ptest/cases/gomspace_checkout_case.py
```diff
@@ -0,0 +1,152 @@
...
+                                                               + str(i) + "_cmd", "true"))
+                                  for i in range(1, 7)]
+        # wait for outputs to be off
+        while (not all(out == False for out in output)) and cycle_no - cycle_no_init < 600:
+            output = [self.str_to_bool(read_state(self, "gomspace.output.output" + str(i)))
+                      for i in range(1, 7)]
+            simulation.flight_controller.write_state(
+                self, "pan.cycle_no", cycle_no + 1)
+            cycle_no = int(read_state(self, "pan.cycle_no"))
+        # wait for outputs to turn on again
```
fatimayousuf, line 112: After this while loop, you should probably print an error if the outputs still haven't turned off after 1 minute

python/ptest/cases/gomspace_checkout_case.py
```diff
@@ -0,0 +1,152 @@
...
+        while (not all(out == True for out in output)) and cycle_no - cycle_no_init < 600:
+            output = [self.str_to_bool(read_state(self, "gomspace.output.output" + str(i)))
+                      for i in range(1, 7)]
+            simulation.flight_controller.write_state(
+                "pan.cycle_no", cycle_no + 1)
+            cycle_no = int(read_state(self, "pan.cycle_no"))
+        # check if finished power cycling
+        for n in range(0, len(power_cycle_output_cmd)):
+            if power_cycle_output_cmd[n] == True:
```
fatimayousuf: I think this should be `if power_cycle_output_cmd[n] == False:` After the power cycle is over, all the outputs should be on again

afoarce: Shouldn't the power cycle output command be false once power cycle finishes, but the outputs are true?

python/ptest/cases/gomspace_checkout_case.py
```diff
@@ -0,0 +1,152 @@
...
+                self, "pan.cycle_no", cycle_no + 1)
+            cycle_no = int(read_state(self, "pan.cycle_no"))
+        # wait for outputs to turn on again
+        while (not all(out == True for out in output)) and cycle_no - cycle_no_init < 600:
+            output = [self.str_to_bool(read_state(self, "gomspace.output.output" + str(i)))
+                      for i in range(1, 7)]
+            simulation.flight_controller.write_state(
+                "pan.cycle_no", cycle_no + 1)
+            cycle_no = int(read_state(self, "pan.cycle_no"))
+        # check if finished power cycling
```
fatimayousuf, line 122: Same thing, you should probably print an error if the outputs still haven't turned on again after 1 minute

fatimayousuf (approved):

tanishqaggarwal (approved):
Looks awesome! I'd merge after implementing @fatimayousuf's recent suggestions.

PR #120, opened 2020-02-08
Dcdc test cases

fatimayousuf:


python/ptest/cases/dcdc_checkout_case.py
```diff
@@ -0,0 +1,100 @@
+from .base import SingleSatOnlyCase
+
+class DCDCCheckoutCase(SingleSatOnlyCase):
+
+    def setup_case_singlesat(self, simulation):
+        simulation.flight_controller.write_state(
```
tanishqaggarwal: Manual state is now 11, sorry.

python/ptest/cases/dcdc_checkout_case.py
```diff
@@ -0,0 +1,100 @@
...
+    def run_case_singlesat(self, simulation):
+        simulation.cycle_no = simulation.flight_controller.read_state("pan.cycle_no")
+
+        adcs_cmd = simulation.flight_controller.read_state("dcdc.ADCSMotor_cmd")
+        sph_cmd = simulation.flight_controller.read_state("dcdc.SpikeDock_cmd")
+        disable_cmd = simulation.flight_controller.read_state("dcdc.disable_cmd")
+        reset_cmd = simulation.flight_controller.read_state("dcdc.reset_cmd")
+        adcs_pin = simulation.flight_controller.read_state("dcdc.ADCSMotor")
+        sph_pin = simulation.flight_controller.read_state("dcdc.SpikeDock")
+        if adcs_cmd==sph_pin and sph_cmd==sph_pin and disable_cmd=="false" and reset_cmd=="false":
```
tanishqaggarwal: if adcs_cmd==adcs_pin

python/ptest/cases/dcdc_checkout_case.py
```diff
@@ -0,0 +1,100 @@
...
+
+    def run_case_singlesat(self, simulation):
+        simulation.cycle_no = simulation.flight_controller.read_state("pan.cycle_no")
+
+        adcs_cmd = simulation.flight_controller.read_state("dcdc.ADCSMotor_cmd")
+        sph_cmd = simulation.flight_controller.read_state("dcdc.SpikeDock_cmd")
+        disable_cmd = simulation.flight_controller.read_state("dcdc.disable_cmd")
+        reset_cmd = simulation.flight_controller.read_state("dcdc.reset_cmd")
+        adcs_pin = simulation.flight_controller.read_state("dcdc.ADCSMotor")
+        sph_pin = simulation.flight_controller.read_state("dcdc.SpikeDock")
```
tanishqaggarwal: The naming is a little confusing here; the "_pin" suffix to me implies that you're reading a pin # off the device, which is not actually what's happening. I suggest "adcs_rd".

python/ptest/cases/dcdc_checkout_case.py
```diff
@@ -0,0 +1,100 @@
...
+        if adcs_cmd==sph_pin and sph_cmd==sph_pin and disable_cmd=="false" and reset_cmd=="false":
+            print("Control task initialized correctly")
+
+        # Try both DCDCs on. Turn on all systems (ADCS motors, prop valves, docking motor). 
+
+        print ("Test case 1: ")
+
+        simulation.flight_controller.write_state("dcdc.ADCSMotor_cmd", "true")
+        simulation.flight_controller.write_state("dcdc.SpikeDock_cmd", "true")
+        adcs_pin = simulation.flight_controller.read_state("dcdc.ADCSMotor")
```
tanishqaggarwal: You should have to cycle in between here using the "cycle.start" state field

python/ptest/cases/dcdc_checkout_case.py
```diff
@@ -0,0 +1,100 @@
...
+        simulation.flight_controller.write_state(
+            "pan.state", 9)  # Manual state
+        self.run_case_singlesat(simulation)
+        print("DCDC cases finished.")
+
+    def run_case_singlesat(self, simulation):
+        simulation.cycle_no = simulation.flight_controller.read_state("pan.cycle_no")
+
+        adcs_cmd = simulation.flight_controller.read_state("dcdc.ADCSMotor_cmd")
+        sph_cmd = simulation.flight_controller.read_state("dcdc.SpikeDock_cmd")
```
tanishqaggarwal: To make this test case prettier maybe you can define class members that make the state fields a property of the class:

```
@property
def adcs_motor_cmd(self): 
    return self.sim.flight_controller.read_state("dcdc.ADCSMotor_cmd")
    
    # setting the values     
    @value.setter 
    def adcs_motor_cmd(self, value): 
        assert(value == True or value == False)
        self.sim.flight_controller.write_state("dcdc.ADCSMotor_cmd", true)
```

Then reading and writing the state field is as simple as doing `x = self.adcs_motor_cmd` and `self.adcs_motor_cmd = "whatever"`. It reduces the clutter in the test file.

tanishqaggarwal: @afoarce I recommend this too for future testcase development. Not sure how applicable it is to Gomspace currently but there's probably some room for prettification there.

python/ptest/cases/dcdc_checkout_case.py
```diff
@@ -0,0 +1,100 @@
...
+        disable_cmd = simulation.flight_controller.read_state("dcdc.disable_cmd")
+        reset_cmd = simulation.flight_controller.read_state("dcdc.reset_cmd")
+        adcs_pin = simulation.flight_controller.read_state("dcdc.ADCSMotor")
+        sph_pin = simulation.flight_controller.read_state("dcdc.SpikeDock")
+        if adcs_cmd==sph_pin and sph_cmd==sph_pin and disable_cmd=="false" and reset_cmd=="false":
+            print("Control task initialized correctly")
+
+        # Try both DCDCs on. Turn on all systems (ADCS motors, prop valves, docking motor). 
+
+        print ("Test case 1: ")
```
tanishqaggarwal: I think it's a good idea to be more descriptive with the test case when printing to the console. No need for the comment on line 23: just put it directly into the print statement! That way you get the best of both worlds: inline documentation of the code, and documentation of the test as it prints to the screen.

tanishqaggarwal (changes requested):
Made some stylistic suggestions. Overall test plan looks good, nice job!

python/ptest/configs/fc_only_teensy.json
```diff
@@ -1,12 +1,12 @@
 {
     "seed" : 2,
-    "sim_duration": 0,
+    "sim_duration": 9000000000000000000000000000000000000000000000000000000000,
```
tanishqaggarwal, line 3: lol

python/ptest/cases/dcdc_checkout_case.py
```diff
@@ -0,0 +1,114 @@
...
+        else:
+            print("Unable to turn ADCS Motor off and SpikeDock on")
+
+        print("Test Case 4: ADCS DCDC off, SPH + Prop DCDC off. Turn on all systems. ")
+
+        self.disable_cmd="true"
+        if self.adcs_rd=="false" and self.sph_rd=="false":
+            self.reset_cmd="true"
+            # Reset takes at least one control cycle to complete
+            self.simulation.flight_controller.write_state("cycle.start", "true")
```
tanishqaggarwal: You can probably make cycle-starting a property or function in the base-level case file (base.py)

tanishqaggarwal (approved):
Can you wait for Andreea's review and for @afoarce to address and merge her PR? Your PR has unrelated Gomspace changes.

Feel free to either implement or not implement the change I suggested.

python/ptest/cases/base.py
```diff
@@ -16,6 +16,9 @@ def setup_case(self, simulation):
     def run_case(self, simulation):
         raise NotImplementedError
 
+    def cycle(self, simulation):
+        simulation.flight_controller.write_state("cycle.start", "true")
```
tanishqaggarwal: You could use `self.sim` instead of `simulation` and this makes the syntax marginally cleaner, since you don't have to pass in simulation as an argument

python/ptest/cases/dcdc_checkout_case.py
```diff
@@ -0,0 +1,114 @@
...
+            print("Control task initialized correctly") 
+
+        print ("Test case 1: Try both DCDCs on. Turn on all systems (ADCS motors, prop valves, docking motor).")
+
+        self.adcs_cmd="true"
+        self.sph_cmd="true"
+
+        if self.adcs_rd=="true" and self.sph_rd=="true":
+            print("Passed")
+        else:
```
afoarce, line 69: Maybe specify which is not on or if both failed

python/ptest/cases/dcdc_checkout_case.py
```diff
@@ -0,0 +1,114 @@
...
+        self.adcs_cmd="true"
+        self.sph_cmd="false"
+        if self.adcs_rd=="true" and self.sph_rd=="false":
+            self.reset_cmd="true"
+            # Reset takes at least one control cycle to complete
+            self.cycle()
+            if (self.adcs_rd == "true" and self.sph_rd == "true"):
+                print("Passed")
+            else:
+                print("Unable to reset pins")
```
afoarce, line 83: could print the values of pins whenever a test fails

afoarce (approved):
Looks good overall, could maybe specify values to user when a test fails.

PR #127, opened 2020-02-26
Sensor models

srutiv:
mag characterization: bias vs frequency, offsets between frequency and bias oscillation, driver noise during ramp-up, noise vs freq (in progress, playing with filters)

---
This change is [Reviewable](https://reviewable.io/reviews/pathfinder-for-autonomous-navigation/psim/127)


kylekrol:
@srutiv Just pinging this to see if we can get this PR cleaned up as well! I think there are a lot of extra commits in here (particularly the geograv stuff). Also, moving the data out of the PR would be great too!

srutiv:
> @srutiv Just pinging this to see if we can get this PR cleaned up as well! I think there are a lot of extra commits in here (particularly the geograv stuff). Also, moving the data out of the PR would be great too!

For sure. Gonna still be adding a few last-minute data processing stuff that Stewart wanted. Will clean up the PR right after.

python/usb_console/configs/radio_keys.json
```diff
@@ -0,0 +1,5 @@
+{
+    "email_username": "pan.ssds.qlocate@gmail.com",
+    "email_password": "OAANisKOOL7373",
+    "imei": 300234030627990
+}
```
shihaocao, line 5: We should probably remove this haha

PR #128, opened 2020-02-28
Mission manager ptest case

tanishqaggarwal:
To run, invoke the command

`python -m ptest.run_simulation -c ptest/configs/fc_only_native.json -t DeploymentToStandby`

from within the psim/python directory. Be sure that you've activated the virtual environment and edited ptest/configs/fc_only_native.json to point to the correct location of FlightSoftware (or, if you're on Windows, use fc_only_teensy.json and then modify the USB port to which the Teensy is connected)

---
This change is [Reviewable](https://reviewable.io/reviews/pathfinder-for-autonomous-navigation/psim/128)


python/ptest/cases/base.py
```diff
@@ -1,91 +1,200 @@
...
+
     @property
-    def run_sim(self):
-        return False
+    def sim_duration(self):
+        return 0
+
+    @property
+    def sim_initial_state(self):
+        return 'startup'
```
tanishqaggarwal, line 103: We should try to wrap these properties up into a struct @nhz2 @kkrol27

MATLAB/sensor_reading.m
```diff
@@ -70,6 +70,11 @@
     end
 end
 
+% This is a bandaid fix to let the DeploymentToStandby ptest case work.
+% The testcase depends on the attitude estimator working so that the satellite can
+% detumble quickly, but the estimator cannot work if the GPS position is NaN because
+% the GPS hasn't yet acquired lock. So this line of code forces GPS lock to exist.
+sensor_readings.position_ecef= position_ecef;
```
nhz2: This should be in an if statement controlled by a variable in const, that way it won't break the matlab sim.

nhz2 (changes requested):

nhz2 (approved):

kylekrol (approved):

kylekrol:
@tanishqaggarwal What's the story here with this PR and the PTest merger into FSW?

tanishqaggarwal:
Closing because the change has moved to FlightSoftware.

PR #129, opened 2020-03-01
gomspace ptest changes

fatimayousuf:


afoarce (approved):

python/ptest/cases/gomspace_checkout_case.py
```diff
@@ -100,7 +100,7 @@ def write_state(self, string_state, state_value):
                                                                + str(i) + "_cmd", "true"))
                                   for i in range(1, 7)]
         # wait for outputs to be off
-        while (not all(out == False for out in output)) and cycle_no - cycle_no_init < 600:
+        while (not all(out == False for out in output)) and cycle_no - cycle_no_init <= 600:
```
tanishqaggarwal: Why did this have to change?

python/ptest/cases/gomspace_checkout_case.py
```diff
@@ -133,29 +133,36 @@ def write_state(self, string_state, state_value):
         if ppt_mode_cmd == ppt_mode_updated:
             print("Could not update pptmode")
 
-        heater_cmd = self.str_to_bool(read_state(self, "gomspace.heater_cmd"))
```
tanishqaggarwal, line 136: Instead of a custom str_to_bool function you could use json.loads? Up to you

tanishqaggarwal: Honestly I think it's worth modifying StateSession to automatically handle bools, since it makes all other code cleaner.

python/ptest/cases/gomspace_checkout_case.py
```diff
@@ -133,29 +133,36 @@ def write_state(self, string_state, state_value):
...
+            cycle_no = int(read_state(self, "pan.cycle_no"))
+            heater_updated=self.str_to_bool(read_state(self, "gomspace.heater"))
+
+        if (heater_init==heater_updated):
+            print("could not update heater")
 
         counter_reset_cmd = self.str_to_bool(read_state(
             self, "gomspace.counter_reset_cmd"))
         counter_reset_cmd_updated = self.str_to_bool(write_state(
-            self, "gomspace.counter_reset_cmd", not counter_reset_cmd))
```
tanishqaggarwal, line 145: Shouldn't you check the counter values to see if the counter(s) resetted?

python/ptest/cases/gomspace_checkout_case.py
```diff
@@ -133,29 +133,36 @@ def write_state(self, string_state, state_value):
...
+            self, "gomspace.counter_reset_cmd", str(not counter_reset_cmd).lower())
         if counter_reset_cmd == counter_reset_cmd_updated:
             print("Could not update counter_reset")
 
         gs_reset_cmd = self.str_to_bool(
             read_state(self, "gomspace.gs_reset_cmd"))
         gs_reset_cmd_updated = self.str_to_bool(write_state(
-            self, "gomspace.gs_reset_cmd", not gs_reset_cmd))
+            self, "gomspace.gs_reset_cmd", str(not gs_reset_cmd).lower())
```
tanishqaggarwal, line 159: Same question here, shouldn't we test that things got reset? Like the PPT mode

python/ptest/cases/gomspace_checkout_case.py
```diff
@@ -133,29 +133,36 @@ def write_state(self, string_state, state_value):
...
         if gs_reset_cmd == gs_reset_cmd_updated:
             print("Could not update gs_reset")
 
         gs_reboot_cmd = self.str_to_bool(
             read_state(self, "gomspace.gs_reboot_cmd"))
         gs_reboot_cmd_updated = self.str_to_bool(write_state(
-            self, "gomspace.gs_reboot_cmd", not gs_reboot_cmd))
+            self, "gomspace.gs_reboot_cmd", str(not gs_reboot_cmd).lower()))
         if gs_reboot_cmd == gs_reboot_cmd_updated:
             print("Could not update gs_reboot")
```
tanishqaggarwal, line 168: How did this play out during testing? Did the flight software disconnect from Teensy like we would expect?

python/ptest/cases/gomspace_checkout_case.py
fatimayousuf: Yeah I know, for some reason the counter values aren't changing on ptest when I call reset. I didn't get to finish looking into why that is the case with the reset command today. I just wanted to fix the write state call in this pr, I was planning on fixing the rest of this test during the next work session?

python/ptest/cases/gomspace_checkout_case.py
fatimayousuf: See previous reply

python/ptest/cases/gomspace_checkout_case.py
fatimayousuf: Ah sorry I forgot to change that back! The power cycling on the EDU sat wasn't working, so I just tried making a bunch of changes with the checkout case in an attempt to get it to work. But the power cycling worked just fine with the other test gomspace, so I'm guessing maybe the EDU sat wasn't fully charged?

python/ptest/cases/gomspace_checkout_case.py
fatimayousuf: nope :(

python/ptest/cases/gomspace_checkout_case.py
tanishqaggarwal: If the test isn't ready to be finalized, let's open the PR later when it is.

PR #136, opened 2020-03-08
Minimum Working Example of New Configuration Files

kylekrol:
Opening this to give a little preview of how the configuration files could work for PSim (and effectively PTest in the future). It's pretty early on and very much so a WIP but let me know of any questions, comments, or concerns. Would rather address things either to prevent wasted efforts on this.

The basic idea is we have a hierarchy of configuration files. A parser starts from the highest directories and walks "towards" the specified relative path. At each step along the way it will generate the updated leader and follower dictionaries overwriting previously set values as it moves down the tree.

In each configuration file, there can be a shared, leader, and follower section. Shared applies to both satellites but is overwritten by leader the leader and follower sections at each level.

The modules refers to each set of initial conditions - i.e. truth initial conditions, sensor initial conditions, et cetera.

Currently, I'd suggest taking a look at `python/psim/config.README.md` and giving this command a go:

    cd python
    python -m psim config --modules truth --path deployment

to see how it will sift through the configuration files.

---
This change is [Reviewable](https://reviewable.io/reviews/pathfinder-for-autonomous-navigation/psim/136)


python/psim/cli/config.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+def main(args):
+    
+    parser = argparse.ArgumentParser(
+        prog='psim config', description='Process PSim configuration files.'
+    )
+    parser.add_argument(
+        '--modules', '-m', nargs='+', required=False, type=str, help='Set of ' +
+        'modules to process configuration files for. Defaults to all modules ' +
```
tanishqaggarwal, line 14: Should say "defaults to all modules _if_ none are specified"

python/psim/cli/config.py
```diff
@@ -0,0 +1,35 @@
...
+        '--output', '-o', type=argparse.FileType('w'), default='-',
+        help='Destination file for the processed configuration file. Defaults' +
+        ' to the stdout.'
+    )
+    parser.add_argument(
+        '--path', '-p', required=True, type=str, help='PSim configuration directory.'
+    )
+    args = parser.parse_args(args)
+
+    leader, follower = config.load_module(args.modules[0], args.path)
```
tanishqaggarwal, line 27: Is it worth inserting  a "TODO" here since only the first module's config is currently getting processed?

How will the argparse interface change to accommodate configs for other kinds of modules?

python/psim/cli/config.py
```diff
@@ -0,0 +1,35 @@
...
+        prog='psim config', description='Process PSim configuration files.'
+    )
+    parser.add_argument(
+        '--modules', '-m', nargs='+', required=False, type=str, help='Set of ' +
+        'modules to process configuration files for. Defaults to all modules ' +
+        'none are specified.'
+    )
+    parser.add_argument(
+        '--output', '-o', type=argparse.FileType('w'), default='-',
+        help='Destination file for the processed configuration file. Defaults' +
```
tanishqaggarwal, line 19: Is this actually defaulting to stdout? I see `-` as the default argument

python/psim/README.md
```diff
@@ -0,0 +1,32 @@
+
```
tanishqaggarwal, line 1: Might be worth including some brief install instructions, e.g. set up venv and install argparse?

python/psim/cli/config.py
```diff
@@ -0,0 +1,35 @@
...
+        ' to the stdout.'
+    )
+    parser.add_argument(
+        '--path', '-p', required=True, type=str, help='PSim configuration directory.'
```
tanishqaggarwal, line 23: I think should be mentioned that the path is relative to psim/config rather than to python/

python/psim/cli/config.py
tanishqaggarwal: Interesting, it seems to work.

tanishqaggarwal (approved):
I just have some questions but this looks pretty solid

python/psim/cli/config.py
kylekrol: Found it online - don't fully understand it lol

python/psim/cli/config.py
kylekrol: Yep this will change!

python/psim/README.md
kylekrol: This will all come in due time. Probably once we can actually startup a MATLAB simulation with the python configuration parser. Although at least on  my machine `argparse` is already installed. Not sure if this is standard or just happened by chance on my machine.

kylekrol:
Sounds good. I will eventually close down this PR and open another one once a full configuration can be passed to the MATLAB simulation using this interface.

kylekrol:
Closing this now and should be opening up another one this weekend!

PR #138, opened 2020-03-09
Install script

nhz2:
I moved the MATLAB C++ code compilation to an `install.m` script. Right now it is optional to run `install.m` because C++ code compilation also happens in `config.m` but in a future PR `config.m` will not do C++ code compilation. 

The reason for this change is 
1. I am adding in [USNavalResearchLaboratory/TrackerComponentLibrary](https://github.com/USNavalResearchLaboratory/TrackerComponentLibrary) which has a huge amount of installation and C++ code compilation, 
2. When running in multiple sims in parallel, if `config.m` writes to shared files by recompiling code that another worker is using, it causes an error.

Ideally the script would somehow use makefiles or something to only recompile C++ code that has changed, but I am not sure how to do this.

---
This change is [Reviewable](https://reviewable.io/reviews/pathfinder-for-autonomous-navigation/psim/138)

### DIFF
--- a/MATLAB/README.md
+++ b/MATLAB/README.md
@@ -373,6 +373,11 @@ One the Matlab simulation side, next getting a complete version of `sensor_readi
 | plot_orbit_error             | Nathan  |          | done                 | NA          | NA          |
 | plot_fancy_animation         | Nathan  |          | done                 | NA          | NA          |
 
+# Installation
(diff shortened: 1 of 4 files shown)


tanishqaggarwal (approved):
Looks good. I'll need to make a change on ptest to check if it's already compiled C++, and if not it should do it.

PR #151, opened 2020-03-19
Drag2dynamics update

srutiv:
Before opening the PR, think: 
- Does this PR break psim, can `main.m` still run, and do all tests in `run_tests.m` pass?
- Is this PR bulletproof? Can I think of no ways to improve this PR beyond its current state?
- Is the PR short enough to ensure a quick but thorough review, but long enough to represent a complete feature? For large state machines, let's try opening PRs for only a few states at a time.

If the answer to any of these questions is "no", refactor your code and open your PR when the answer to both questions is "yes".

--------------------------

# PR Title

Fixes #. (Optional; sometimes PRs don't have an associated ticket)

### Summary of changes
- 
- 
- 

### Ptest Effects
What effect will this have on the ptest interface? Don't use uint64 because uint64 is not python compatible. Be careful when changing sensor readings or actuator commands because these are directly sent to and recieved from the flight computer.

### Testing
Such as your unit and functional testing. The testing should be at a level appropriate to demonstrate that your feature is sufficiently tested.
For new MATLAB functions, added a simple test script to `MATLAB/test` directory and make sure it is called in `MATLAB/run_tests.m`. Plotting functions and functions with IO or other effects should not be tested in `MATLAB/run_tests.m`. Unit tests should be fast, deterministic and silent if they pass, and give lots of debug info only when they fail.

If the functionality can only be tested with a long sim, give instructions for how to do those sims and get the plots. If needed, add a new plotting function to `MATLAB/plot` to help others reproduce your results.

If your update does not require significant testing, argue why, or if you're deferring testing to another PR, create an issue ticket for the deferral and link it.

### Constants
If your PR creates any constants, list them here and indicate whether or not you've added them to the recurring constants review.
If you change or add any fields in `const`, `main_state`, sensor readings or actuator commands make sure to document that change in `MATLAB/README.md`.

### Documentation Evidence
Post to ReadTheDocs, or
- Argue that your inline documentation is sufficient.
- Create a issue ticket deferring the documentation task.


kylekrol:
@srutiv If you could fill in a bit of the PR template (especially summary of changes, documenation, etc.) that'd be helpful for future reference!

.github/PULL_REQUEST_TEMPLATES/pull_request_template.md
```diff
@@ -0,0 +1,43 @@
...
+| Constant                 | Location             | Added to review?   |
+|--------------------------|----------------------|--------------------|
+|                          |                      | :white_check_mark: |
+|                          |                      | :x: : Why not?     |
+
+
+### Documentation Evidence
+Post to ReadTheDocs, or
+- Argue that your inline documentation is sufficient.
+- Create a issue ticket deferring the documentation task.
```
kylekrol, line 43: I think this file somehow ended up in this PR and shouldn't be here?

CODEOWNERS
```diff
@@ -0,0 +1,16 @@
...
+
+python/*                  tanishqaggarwal
+python/gsw/* fatimayousuf tanishqaggarwal
+
+include/*      kkrol27
+src/*          kkrol27
+test/*         kkrol27
+platformio.ini kkrol27
+Dockerfile/*   kkrol27
+tools/*        kkrol27
```
kylekrol, line 16: Same thing with this one - don't think it should be here.

MATLAB/config.m
```diff
@@ -7,6 +7,16 @@
...
+[filepath, name, ext] = fileparts(mfilename("fullpath"));
+addpath(strcat(filepath, '/utl'));
+addpath(strcat(filepath, '/environmental_models'));
+addpath(strcat(filepath, '/environmental_models/helper_functions'));
+addpath(strcat(filepath, '/plot'));
+addpath(strcat(filepath, '/plot/czml_helpers'));
+addpath(strcat(filepath, '/adcs'));
+addpath(strcat(filepath, '/orbit_estimation'));
+addpath(strcat(filepath, '/test'));
 setup_path()
```
kylekrol, line 20: Are these changes also not supposed to be here? I think `setup_path` takes care of it now!

MATLAB/dynamics_update.m
```diff
@@ -82,16 +82,21 @@
         quat_body_eci=y(7:10);
         quat_eci_body= utl_quat_conj(quat_body_eci);
         quat_body_ecef= utl_quat_cross_mult(quat_body_eci,quat_eci_ecef);
-        pos_eci=y(1:3);
+        pos_eci=y(1:3); 
+        vel_eci=y(4:6);
         pos_ecef=utl_rotateframe(quat_ecef_eci,pos_eci);
+        vel_ecef = utl_rotateframe(quat_ecef_eci,vel_eci);
```
kylekrol, line 88: Unsure exactly how we decided on doing the velocity conversion, but we may not be accounting for the velocity component due to ECEF's rotation in ECI (something like `cross(w, r)`). If we don't want to account for this term we should call `vel_ecef` something different.

MATLAB/dynamics_update.m
```diff
@@ -82,16 +82,21 @@
...
         [g_ecef,~,G_ecef]= env_gravity(t,pos_ecef);
         g_eci=utl_rotateframe(quat_eci_ecef,g_ecef);
         %TODO add thruster firings forces and torques
         %TODO add drag and solar pressure forces
-        dydt(4:6)= g_eci;
+        [F_envdrag_ecef] = env_atmospheric_drag(pos_ecef,vel_ecef,quat_body_ecef); %differential drag in ecef
+        F_envdrag_eci = utl_rotateframe(quat_eci_ecef,F_envdrag_ecef);
+        dydt(4:6)= g_eci + F_envdrag_eci./(const.MASS);
```
kylekrol, line 95: Is the call to `env_atmospheric _drag` in particular an expensive function?

kylekrol (changes requested):
I'm pretty sure there are some files in the diff that got caught up in the PR because of a rebase a while ago. It may be easier to start a new branch of master, add in your changes again, and open a new PR to get rid of them.

Other than that, I just had the one question about the `vel_ecef` value and that's about it!

MATLAB/dynamics_update.m
nhz2: You can get earth's rotation rate from `const.earth_rate_ecef` Then add something like 
```suggestion
        vel_ecef = utl_rotateframe(quat_ecef_eci,vel_eci);
        vel_ecef= vel_ecef - cross(const.earth_rate_ecef,pos_ecef);
```

.github/PULL_REQUEST_TEMPLATES/pull_request_template.md
srutiv: very strange. I just started a new branch off of master, and made the required changes for this pull request, but somehow other files and commits still keep getting looped in.

PR #154, opened 2020-03-23
Updated to new lin library

kylekrol:
# Updated `lin` Library

### Summary of changes
- Updated the `lin` library to allow for:
  - Support for C++ style implicit casting in arithmetic operations (e.g. you can add a double vector to a float vector and it will automatically deduce the return type to be a double vector).
  - Restructured modular architecture to help improve compile times if only the "modules" needed by a particular translation unit are included (most of the time only `#include <lin/core.hpp>` is needed unless you specifically need references, generators, factorizations, etc.)
  - Has no affect to PAN specific code but now uses GTest.
  - Can be used as a platformio library dependency so we won't need to clone submodules recursively anymore (could help with CI times as well).
- Updated the GNC constants file to take advantage of the implicit casting and reworked the includes.

This PR will need to have a sister PR in flight software. However, all changes should be backwards compatible because the new `lin` library still has a top level `lin.hpp` file that includes all modules.

### Ptest Effects
NA

### Testing
Passing local tests (and presumably CI).

### Constants
NA

### Documentation Evidence
NA


include/gnc_constants.hpp
```diff
@@ -70,15 +70,11 @@ constexpr static lin::Vector3d earth_precession_rate = 1.0e-11L * lin::Vector3d(
 
 /** Approximate precssion rate of Earth's Axis in radians per second (float
  *  version). */
-constexpr static lin::Vector3f earth_precession_rate_f = {
-  static_cast<float>(earth_precession_rate(0)),
-  static_cast<float>(earth_precession_rate(1)),
-  static_cast<float>(earth_precession_rate(2))
-};
+constexpr static lin::Vector3f earth_precession_rate_f = earth_precession_rate;
```
kkrol27: Just to point it out, this is an example of how we'll be able to take advantage of implicit casts.

kylekrol, line 73: Just to point it out, this is an example of how we'll be able to take advantage of implicit casts.

include/gnc_constants.hpp
```diff
@@ -30,13 +30,13 @@ namespace constant {
 constexpr static double pi = 3.141592653589793L;
 
 /** Good old pi (float version). */
-constexpr static float pi_f = static_cast<float>(pi);
+constexpr static float pi_f = pi;
```
tanishqaggarwal, line 33: Why was this `static_cast`s removed?

src/gnc_attitude_controller.cpp
```diff
@@ -13,6 +13,8 @@
 #include <gnc_attitude_controller.hpp>
 #include <gnc_constants.hpp>
 
+#include <lin/generators/constants.hpp>
+
```
tanishqaggarwal, line 17: Why is this necessary?

tanishqaggarwal (approved):

nhz2 (approved):

src/gnc_attitude_controller.cpp
kylekrol: To give us the `lin::nans<...>(...)`

include/gnc_constants.hpp
kylekrol: And because it's verbose. We all know an assignment to a float will perform the cast anyway

src/gnc_attitude_controller.cpp
kylekrol: I'm going to be working on throwing some documentation together - but the code I think is much much easier to understand from the `*.hpp` files alone. I moved a lot of the BS into the `*.inl` files which people in general shouldn't have to look at.

PR #156, opened 2020-03-25
Added GNC Constants Tracking and Slight Restructuring

### LINKED ISSUE #139: Constants Read/Write Interface and Tracking

kylekrol:
# Added GNC Constants Tracking and Slight Restructuring

Fixes #139.

### Summary of changes
- Transitioned to Github Actions
- Updated the build system to rely on a virtual environment install of PIO (see `requirements.txt`)
- Added a small autocoder to produce the `include/gnc/constants.hpp` and `src/gnc_constants.cpp` files
- Added a GNC constants tracking mechanism that piggybacks off of flight software
- The idea here would be the constants tracker in flight software is still the one source of truth in terms of what constants exist and what there values in initialized to

### Ptest Effects
NA

### Testing
All changes were simply cosmetic. Testing infrastructure was only refactored.

### Constants
The value of each constant remained the same but this is the tracking infrastructure.

### Documentation Evidence
See inline documentation and the updated README. ~~More documentation, however, should be written about the constants format and autocoder.~~


include/gnc/constants.hpp
```diff
@@ -0,0 +1,54 @@
...
+#include <lin/core/vector/vector.hpp>
+
+namespace gnc {
+namespace constant {
+
+GNC_TRACKED_CONSTANT(constexpr static double, pi, 3.141592653589793);
+
+GNC_TRACKED_CONSTANT(constexpr static double, two_pi, 2.0 * pi);
+
+extern unsigned short init_gps_week_number;
```
nhz2, line 20: Why are some of these extern and some constexpr static?

tools/constants_generator.py
```diff
@@ -0,0 +1,86 @@
...
+
+'''
+CPP_FOOTER='''\
+}  // namespace constant
+}  // namespace gnc
+
+'''
+
+
+def generate(csv_file, hpp_file, cpp_file):
```
nhz2, line 51: Add a doc string about what this function does.

include/gnc/constants.hpp
tanishqaggarwal: I think some are extern so that their values can be controlled by FSW (and therefore by the ground, during mission operation.)

tanishqaggarwal (approved):
Did you get to check if FSW catches all the constants?

kylekrol:
> Did you get to check if FSW catches all the constants?

Did not - although I can go ahead and run it to see. I will tie up those loose ends when integrating the changes into flight software!

PR #158, opened 2020-03-26
Add CXX Attitude Controller

### LINKED ISSUE #82: Add Attitude Controller Function

kylekrol:
# CXX Attitude Controller

Fixes #82.

### Summary of changes

- Added the CXX pointing controller interface and implementation.
- Added necessary and additional, helpful constants.

### Ptest Effects

NA.

### Testing

This will be deferred to a future PR. I am going to opening an issue and working on replacing our current MATLAB controller with this CXX implementation and getting a HOOTL case with attitude control hopefully running soon.

### Constants

MATLAB constants weren't touched.

As for the CXX constants, we will need to come up with some location to document the constants and make it very clear to flight software what changing some of the externs' values will do - i.e. ~~right now to configure the pointing controller flight software would, in theory, change the controller's gains and the satellite's moment of inertia matrix once we dock~~ we may make edits to gains, moment intertia measurements, etc from the ground whether this be in the middle of the mission due to off nominal behavior or once we have docked and want to enable ADCS again.

### Documentation Evidence

See inline documentation for the time being.

### DIFF
--- a/include/gnc/attitude_controller.hpp
+++ b/include/gnc/attitude_controller.hpp
@@ -64,19 +64,19 @@ struct PointingControllerState {
  *  objective information) as NaNs. See member documentation for more details.
  *  */
 struct PointingControllerData {
-  /** Desired state of the primary pointing objective. This needs to be a unit
-   *  vector. */
+  /** Desired state of the primary pointing objective in the body frame. This
+   *  needs to be a unit vector. */
   lin::Vector3f primary_desired;
-  /** Current state of the primary pointing objective. This needs to be a unit
-   *  vector. */
+  /** Current state of the primary pointing objective in the body frame. This
+   *  needs to be a unit vector. */
   lin::Vector3f primary_current;
-  /** Desired state of the secondary pointing objective. This needs to be a unit
-   *  vector. */
+  /** Desired state of the secondary pointing objective in the body frame. This
+   *  needs to be a unit vector. */
   lin::Vector3f secondary_desired;
-  /** Current state of the secondary pointing objective. This needs to be a unit
-   *  vector. */
+  /** Current state of the secondary pointing objective in the body frame. This
+   *  needs to be a unit vector. */
   lin::Vector3f secondary_current;
-  /** Angular rate of the wheels in the body frame (units of radians/s) */
+  /** Angular rate of the wheels in the body frame (units of radians/s). */
   lin::Vector3f w_wheels;
   /** Angular rate of the satellite in the body frame (units of radians/s). */
   lin::Vector3f w_sat;
@@ -90,9 +90,9 @@ struct PointingControllerData {
  *  Essentially serves as the outputs to the control_pointing function. See
  *  member documentation for more information. */
 struct PointingActuation {
-  /** Magnetourquer actuation command in the body frame (units Am^2) */
+  /** Magnetourquer actuation command in the body frame (units Am^2). */
   lin::Vector3f mtr_body_cmd;
-  /** Reaction wheel torque command in the body frame (units Nm) */
+  /** Reaction wheel torque command in the body frame (units Nm). */
   lin::Vector3f rwa_body_cmd;
   /** Defaults everything's value to NaN. */
   PointingActuation();
--- a/include/gnc/attitude_controller.hpp
+++ b/include/gnc/attitude_controller.hpp
@@ -104,8 +104,17 @@ struct PointingActuation {
  *  @param[out]   actuation Controller's outputs.
  *  Calculates wheel and MTR actuations to point to spacecraft according to the
  *  passed objectives. */
+#ifndef MEX
 void control_pointing(PointingControllerState &state,
     PointingControllerData const &data, PointingActuation &actuation);
+#else
+/** @fn mex_control_pointing
+ *  Mex function entrypoint for MATLAB use. It decouples the pointing controller
+ *  from the global controller gains and satellite's moment of inertia value. */
+void mex_control_pointing(control_pointing(PointingControllerState &state,
+    PointingControllerData const &data, PointingActuation &actuation, float Kp,
+    float Kd, lin::Matrix3x3f const &J);
+#endif
 
 }  // namespace gnc
 
--- a/include/gnc/attitude_controller.hpp
+++ b/include/gnc/attitude_controller.hpp
@@ -5,6 +5,7 @@
 #ifndef GNC_ATTITUDE_CONTROLLER_HPP_
 #define GNC_ATTITUDE_CONTROLLER_HPP_
 
+#include "config.hpp"
 #include "containers.hpp"
 
 #include <lin/core.hpp>
@@ -48,6 +49,24 @@ struct DetumbleActuation {
  *  @param[out]   actuation Actuation output. */
 void control_detumble(DetumbleControllerState &state,
     DetumbleControllerData const &data, DetumbleActuation &actuation);
+
+struct PointingControllerState {
+  /** Defaults everything's value to NaN. */
+  PointingControllerState();
+};
+
+struct PointingControllerData {
+  /** Defaults everything's value to NaN. */
+  PointingControllerData();
+};
+
+struct PointingActuation {
+  /** Defaults everything's value to NaN. */
+  PointingActuation();
+};
+
+void control_pointing(PointingControllerState &state,
+    PointingControllerData const &data, PointingActuation &actuation);
 
 }  // namespace gnc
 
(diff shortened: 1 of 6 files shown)


tools/constants.csv
```diff
@@ -12,6 +14,9 @@ false,double,earth_perihelion_time,-6287982.0
 false,double,earth_period,365.256363004 * 24.0 * 60.0 * 60.0
 false,lin::Vector4d,q_eci_perifocal,0.127456632677880,-0.158124280715206,0.762378784011859,-0.614434787689722
 false,double,b_noise_floor,0.0
-false,double,max_mtr_moment,0.113337 / 2.0
-false,lin::Matrix3x3f,JB_single_sat,0.0333f,0.0f,0.0f,0.0f,0.0333f,0.0f,0.0f,0.0f,0.0067f
-false,lin::Matrix3x3f,JB_docked_sats,0.0333f,0.0f,0.0f,0.0f,0.0333f,0.0f,0.0f,0.0f,0.0067f
\ No newline at end of file
+false,float,max_mtr_moment,0.113337f / 2.0f
+true,lin::Matrix3x3f,J_sat,0.0333f,0.0f,0.0f,0.0f,0.0333f,0.0f,0.0f,0.0f,0.0067f
```
kylekrol: I just wanted to make note of this here really quick, I think that FSW should be concerned about changing up our moment of inertia externally. Seems to make more sense to me at least (same with the gains).

tanishqaggarwal (approved):
Looks good to me

tools/constants.csv
tanishqaggarwal: Do you mean that FSW should take care of creating a system for changing the values if required? Sure, if you could open a ticket we could track that work?

kylekrol: Yep sounds good - was going to open one up!

src/gnc_attitude_controller.cpp
```diff
@@ -54,5 +58,120 @@ void control_detumble(DetumbleControllerState &state,
...
+      std::isnan(data.secondary_desired(0))) {
+    single_objective_error(data.primary_current, data.primary_desired, err);
+  }
+  // Secondary objective is specified
+  else {
+    // Check is the primary and secondary objectives are too near
+    float thresh = std::cos(10.0f * constant::pi_f / 180.0f);  // within 10 deg.
+    if (std::abs(lin::dot(data.secondary_current, data.primary_current)) < thresh ||
+        std::abs(lin::dot(data.secondary_desired, data.primary_desired)) < thresh)
+      single_objective_error(data.primary_current, data.primary_desired, err);
```
kylekrol, line 147: Sorry about that. Forgot to put the 10 degree threshold in before!

kylekrol:
> Looks good to me

I'll let Nathan take a look as well but go ahead and make that issue in flight software.

tools/constants.csv
nhz2: I'm not sure about using global variables to control the pointing controller behavior. Shouldn't the moment of inertia be owned by the estimator as part of its internal state, then set or get with functions? Also when running a sim with two sats on the same computer, will the global variable be shared by both leader and follower?

include/gnc/constants.hpp
```diff
@@ -41,11 +45,17 @@ GNC_TRACKED_CONSTANT(constexpr static lin::Vector4d, q_eci_perifocal, 0.12745663
 
 GNC_TRACKED_CONSTANT(constexpr static double, b_noise_floor, 0.0);
 
-GNC_TRACKED_CONSTANT(constexpr static double, max_mtr_moment, 0.113337 / 2.0);
+GNC_TRACKED_CONSTANT(constexpr static float, max_mtr_moment, 0.113337f / 2.0f);
+
+extern lin::Matrix3x3f J_sat;
```
nhz2, line 62: If these are supposed to actually change in normal mission ops, then they should be part of the estimator or controller interface and state (and probably the MATLAB state and interface) instead of just C++ global variables that are "constant" but not really.

nhz2: I feel like having the controller and estimator state and interface in global variables will just get messy and make testing, debugging, and documentation a lot harder. One quick fix is to just create a big class called Context that manages all of these shared parameters, or they could just go into the state field registry (maybe have a control task that manages the shared gnc parameters?). But ideally if the parameters aren't really shared by lots of different things they should be part of the specific controller or estimators internal state and interface.

nhz2 (changes requested):

include/gnc/constants.hpp
kylekrol: Here I was just following the paradigm we have set in flight software. We have a large set of values that we call "constants" that normally live in state fields and are sometimes editable from the ground. In the case of the controllers, I just pushed any "input" that didn't change to frequently out into global constants so we don't have to recopy the values on each call to the controller each time. Furthermore, something like the moment of inertia will definitely be used in the attitude estimator as well except we have no reason for them to be working off of different values.

kylekrol: I'd take a look at the issue linked above from FSW talking about wrapping these editable constants with state fields in flight software. Obviously, it'd be best to make them state fields anyway but we are in different repos so here we are haha And lastly, we could make a context class but then we're never going to have more than one so we could just throw it in a namespace no?

include/gnc/attitude_controller.hpp
```diff
@@ -35,20 +36,77 @@ struct DetumbleControllerData {
...
+};
+
+/** @struct PointingControllerData
+ *  Essentially serves as the inputs to the control_pointing function. Note,
+ *  it is permissable to leave some members (like the secondary pointing
+ *  objective information) as NaNs. See member documentation for more details.
+ *  */
+struct PointingControllerData {
+  /** Desired state of the primary pointing objective. This needs to be a unit
+   *  vector. */
```
nhz2: Document the frame of all of these vectors.

nhz2 (changes requested):

include/gnc/attitude_controller.hpp
kylekrol: Oops yep that definitely needs to be in there.

include/gnc/constants.hpp
stewartaslan: @nhz2 since we're not planning to modify these things under normal operations I don't think it makes sense to pass them all to the controller on every function call. It's simpler to have them defined as constants and it makes sense to retain the ability to edit them in some off-nominal case if we need to.

kylekrol: If we don't want to worry about the docked case and rely on edits from the ground to change the controller state then I agree with that. Our current mission plan just turns off ADCS when we dock and wait for the ground anyway haha I misspoke in the description above then.

stewartaslan (approved):
All looks good to me.

include/gnc/constants.hpp
nhz2: Ok if they are only getting changed from ground in some kind of debug situation thats fine, just realize that it will be very very hard and annoying to test changing any C++ global variables in a MATLAB sim because they will be shared by the entire MATLAB context. So if you are fine not testing the behavior of changing those parameters during a MATLAB sim, and forcing both leader and follower to have the same values, then they can stay as globals.

kylekrol:
@nhz2 @stewartaslan @tanishqaggarwal 

So just a quick summary of the changes I've made to the interface and why.

For the sake of FSW, the original `extern` constant scheme still seems to make sense because those values will only ever change as the result of an explicit radio uplink/command from the ground - not autonomously. As such, it makes sense to leave them as "constants" and wrap them in state fields.

From the MATLAB perspective, we need to be able to pass all of the constants into the MEX function. Otherwise, setting the controller gain, for example, in MATLAB wouldn't change our C++ controller and that's an issue. We need full control for our mex functions.

To hopefully accommodate both to everyone's satisfaction, I've created two separate interfaces in which the code compilers either an `extern mex_control_pointing` function when the macro `MEX` is defined, or an `extern control_pointing` function. The former takes all required constants as an input (controller gains and moment of inertia) and the latter will be used by FSW and pulls from the GNC "constants".

I decided to create two interfaces as opposed to defaulted function arguments to ensure that FSW uses the function __as intended__ and doesn't deviate from the defined interface.

In passing, I've also added in something constants to convert from radians to degrees and vice versa.

nhz2 (approved):
Looks good, before this goes into flight software though make sure to update [https://pan-software.readthedocs.io/en/latest/flight_software/components.html](https://pan-software.readthedocs.io/en/latest/flight_software/components.html) to talk about how some special GNC state fields refer to global variables and are not made and owned by control tasks on construction like all the other state fields. So constructing two of the same type of control task with separate state field registries could lead to unexpected behavior.

kylekrol:
Tracking that in this issue: https://github.com/pathfinder-for-autonomous-navigation/FlightSoftware/issues/287

Issue #160, opened 2020-03-27
Create MEX Functions for the Pointing Controller


kylekrol:
Create a MEX wrapper for the attitude pointing controller.

The current pointing controller interface  is defined in `include/gnc/attitude_controller.hpp` and is given below:

```
struct PointingControllerState {
  /** Defaults everything's value to NaN. */
  PointingControllerState();
};

struct PointingControllerData {
  lin::Vector3f primary_desired;
  lin::Vector3f primary_current;
  lin::Vector3f secondary_desired;
  lin::Vector3f secondary_current;
  lin::Vector3f w_wheels;
  lin::Vector3f w_sat;
  lin::Vector3f b;
  /** Defaults everything's value to NaN. */
  PointingControllerData();
};

struct PointingActuation {
  lin::Vector3f mtr_body_cmd;
  lin::Vector3f rwa_body_cmd;
  /** Defaults everything's value to NaN. */
  PointingActuation();
};

void mex_control_pointing(control_pointing(PointingControllerState &state,
    PointingControllerData const &data, PointingActuation &actuation, float Kp,
    float Kd, lin::Matrix3x3f const &J);
```

We'll need this to be wrapped in a MATLAB function with the following interface:

```
function pointer = make_mex_pointer()
pointer = struct()
pointer.control = @control;
end

function [mtr_body_cmd, rwa_body_cmd] = control(...
    primary_desired, primary_current,...
    secondary_desired, secondary_current,...
    w_wheels, w_body, b_body, Kp, Kd, J) ... end
```

kylekrol:
Closing in favor of C++ psim.

PR #166, opened 2020-03-29
Updated GNC Utility Functions and Attitude Estimator

### LINKED ISSUE #159: Update CXX Attitude Estimator to Work Assuming Perfect Sensors
If we want to get some version of flight software up and running in a HOOTL environment with the new attitude controller, we'll need to make some small changes to the attitude estimator interface. Most notably, the finite difference method to determine `w_body` should be removed because we will be relying on filtered gyroscope measurements in the future.

Obviously, this implementation will be removed in the future.

### LINKED ISSUE #161: Refactor the Triad Utility Function
Currently, the arguments to the triad utility function are named assuming we are only using it for attitude determination given sun and magnetic field vectors. This isn't actually the case, so it'd be nice if we could refactor it to take more general arguments.

### LINKED ISSUE #163: Improve NaN Support in CXX GNC Utility Functions

kylekrol:
# Updated GNC Utility Functions and Attitude Estimator

Fixes #159.
Fixes #161.
Fixes #163.

### Summary of changes
- Put in an explicit threshold on how "close" the triad vector pairs can be before an NaNs quaternion is returned (currently set to `1.0` degrees).
- Add asserts to enforce invariants on GNC utility functions.
- Update GNC utility function documentation to comment on NaN handling.
- Add in `w_sat` as an input to the orbit estimator and remove the finite differencing of the attitude quaternion. We currently are just passing the measured omega through as the "estimate", however, this will get replaced by a full filter in the future.

### Ptest Effects

Not applicable at the moment. Although, with this merged, work can commence on [this](https://github.com/pathfinder-for-autonomous-navigation/FlightSoftware/issues/287).

### Testing
Deferring testing for [mex functions](https://github.com/pathfinder-for-autonomous-navigation/psim/issues/160) and [fsw integration](https://github.com/pathfinder-for-autonomous-navigation/FlightSoftware/issues/287).

### Constants

No change.

### Documentation Evidence

See inline documentation.


include/gnc/utilities.hpp
```diff
@@ -15,60 +20,93 @@ namespace gnc {
...
- * 
- *  REQUIRES: N_* and B_* vectors must be unit vectors. */
+ *  @param[in]  R1 First known vector in a known frame.
+ *  @param[in]  R2 Second known vector in a known frame.
+ *  @param[in]  r1 First measurement vector in an unknown frame.
+ *  @param[in]  r2 Second measurement vector in an unknown frame.
+ *  @param[out] q  Quaternion transforming from the unknown to known frame.
+ *  Given a set of unit vector from a known frame and then unknown frame, this
+ *  function calculates the rotation required to transform from the latter to
+ *  the former. If R1 and R2, or r1 and r2 are within a degree of one another,
```
nhz2, line 88: I would add that q will rotate the first vector exactly, and the second vector as best it can.

kylekrol: Oh shoot yep - will do

nhz2: Also that q is quat_unknown_known so it should rotate from known frame to unknown frame. Also, there are no MATLAB tests for triad, its copied from Justin's Simulink, and the docs in `utl_triad.m` aren't very helpful. #167

src/gnc_attitude_estimator.cpp
```diff
@@ -0,0 +1,74 @@
...
+static void estimate_q_body_eci(double t, lin::Vector3d const &r_ecef,
+    lin::Vector3f const &s_body, lin::Vector3f const &b_body,
+    lin::Vector4f &q_body_eci) {
+  lin::Vector3f s_eci, b_eci;
+
+  // Default to NaNs
+  q_body_eci = lin::nans<lin::Vector4f>();
+
+  // Determine the expected magnetic field
+  env::magnetic_field(t, lin::Vector3f(r_ecef), b_eci);
```
nhz2, line 38: I think this should be b_ecef

include/gnc/utilities.hpp
kylekrol: Oh wait really? Welp, I may need to rework some documentation in other places as well. Really not sure where I got that from. Thanks for keeping an eye out!

nhz2 (changes requested):
The magnetic field in ECEF should be rotated into the ECI frame for triad.

src/gnc_attitude_estimator.cpp
kylekrol: Oh my yes indeed.

kylekrol: Alright this should be fixed.

include/gnc/utilities.hpp
kylekrol: And greatly improved the documentation in CXX at least for triad.

kylekrol:
Oh shoot I'm sorry - this is the wrong branch

PR #168, opened 2020-03-29
Updated GNC Utility Functions and Attitude Estimator

### LINKED ISSUE #159: Update CXX Attitude Estimator to Work Assuming Perfect Sensors

### LINKED ISSUE #161: Refactor the Triad Utility Function

### LINKED ISSUE #163: Improve NaN Support in CXX GNC Utility Functions

kylekrol:
Changes here shouldn't be too terribly different from the PR that I just opened. Had just forgot that I merged two sets of changes together!

# Updated GNC Utility Functions and Attitude Estimator

Fixes #159.
Fixes #161.
Relates to #163.

### Summary of changes
- Put in an explicit threshold on how "close" the triad vector pairs can be before an NaNs quaternion is returned (currently set to `1.0` degrees).
- Add asserts to enforce invariants on GNC utility functions.
- Update GNC utility function documentation to comment on NaN handling.
- Add in `w_sat` as an input to the orbit estimator and remove the finite differencing of the attitude quaternion. We currently are just passing the measured omega through as the "estimate", however, this will get replaced by a full filter in the future.

### Ptest Effects

Not applicable at the moment. Although, with this merged, work can commence on [this](https://github.com/pathfinder-for-autonomous-navigation/FlightSoftware/issues/287).

### Testing
Deferring testing for [mex functions](https://github.com/pathfinder-for-autonomous-navigation/psim/issues/160) and [fsw integration](https://github.com/pathfinder-for-autonomous-navigation/FlightSoftware/issues/287).

### Constants

No change.

### Documentation Evidence

See inline documentation.


include/gnc/config.hpp
```diff
@@ -27,6 +27,9 @@
  *  used with floating point values. */
```
tanishqaggarwal, line 27: When is GNC_NO_ASSERT defined? I'm uncomfortable with HOOTLs suddenly failing because over time, numerical errors cause these assertions to fail. A quieter error would be better. Is that possible?

tanishqaggarwal (approved):
Looks good, looks mostly like cosmetic changes except for the slight change in interface.

include/gnc/config.hpp
nhz2: Yeah, I agree, also in the attitude estimator, assertions that vectors are normalized will fail if the vector has NANs. I like the use of asserts to though, they might make debugging weird numerical issues a lot easier, maybe though there could be an option for printing a warning and setting output to NAN or something instead of crashing?

src/gnc_attitude_estimator.cpp
```diff
@@ -0,0 +1,78 @@
...
+  // Ensure the measured and expected magnetic field are large enough
+  float thresh = constant::b_noise_floor * constant::b_noise_floor;
+  if ((lin::fro(b_eci) < thresh) || (lin::fro(b_body) < thresh))
+    return;
+
+  // Determine the expected sun vector
+  env::sun_vector(t, s_eci);
+
+  // Ensure the sun vectors are unit vectors
+  GNC_ASSERT_NORMALIZED(s_body);
```
nhz2, line 53: This will fail if the sun sensors aren't working, because s_body is NAN be default. You could add a check like on line 46 that returns if any of the inputs are not finite.

nhz2 (changes requested):

src/gnc_attitude_estimator.cpp
kylekrol: Line 70 I think checks that the sun vector is finite so it should be fine right?

```
// Ensure the magnetic field vector, sun vector, ECEF position, and time were
// given
if (std::isnan(data.b_body(0)) || std::isnan(data.s_body(0)) ||
    std::isnan(data.r_ecef(0)) || std::isnan(data.t))
  return;
```

`estimate_q_body_eci` is a static function local only to this translation unit.

include/gnc/config.hpp
kylekrol: I sort of assumed that we'd disable the asserts in FSW; the GNC algorithms should be verified enough by then. Furthermore, I do share the concern about HOOTLs failing over time due to numerical errors. However, I've just placed asserts where this shouldn't be an issue and where strict invariants need to be enforced. For example, all vectors being passed in to triad should have been very recently normalized otherwise we __are__ doing something very wrong, built up numerical errors or not.
On using NaNs to signal, while yes I agree it's effective I don't think we should rely on letting NaNs propegate wildly through are flight code. This would give us just as much trouble or possible more trouble than assertions because it's a quiet, unsignaled failure. I've been very careful to explicitly define where it's possible for NaNs to occure and always place a check immediately afterwards (basically using NaNs as an exit code).
Lastly, I do like the idea of printing something with asserts potentially further down the line. However, I'd rather get changes to `GNC_ASSERT` tackled at a later date because I don't forsee it being too much of an issue. If it is, we can always revisit given it's one line of code.

kkrol27: Using NaNs to signal errors works but it can almost give us just as much trouble as assertions because it's very difficult to track the origin of the problem.

kkrol27: I do like the idea of printing something, although given our wide use of platforms including mex functions, teensies, desktop, etc. handling the different forms of IO turns into a decent amount of work. I'd be in favor of leaving it as is for now and if we run into issues we can look into better definitions of `GNC_ASSERT`?

src/gnc_attitude_estimator.cpp
nhz2: Yeah missed that.

nhz2 (approved):

PR #169, opened 2020-03-29
Orbit Class

nhz2:
# Orbit Class

Related to #78

### Summary of changes

Just a layout of the methods of the class so far. The goal is to create a class to handle most of the logic related to storing, serializing, propagating, and checking validity of an orbit. This class will be used by the next version of the orbit estimator, and if it works out well probably become a C++ class in flight-software. Another goal is to specify, control, and minimize the number of calls to the gravity function, because one call could take about 1ms on a teensy.

### Ptest Effects
I am adding a `+Orbit` namespace/package. To call functions in this package use for example
`eng.Orbit.valid(myorbit)` where `myorbit` is a `dict` of an Orbit.

### Testing
I'll be adding unit testing for the class soon.

### Constants
If your PR creates any constants, list them here and indicate whether or not you've added them to the recurring constants review.
If you change or add any fields in `const`, `main_state`, sensor readings or actuator commands make sure to document that change in `MATLAB/README.md`.

### Documentation Evidence
TODO make a ReadTheDocs post about this before merging.

### DIFF
--- a/MATLAB/orbit_estimation/+Orbit/abortpropagator.m
+++ b/MATLAB/orbit_estimation/+Orbit/abortpropagator.m
@@ -1,7 +1,0 @@
-function [self] = abortpropagator(self)
-%ABORTPROPAGATOR Aborts the propagator,
-%   If self is propagating this restores self to before
-%   startpropagator was called. If self is not propagating, this does
-%   nothing.
-end
-
--- a/MATLAB/orbit_estimation/+Orbit/abortpropagator.m
+++ b/MATLAB/orbit_estimation/+Orbit/abortpropagator.m
@@ -1,0 +1,7 @@
+function [self] = abortpropagator(self)
+%ABORTPROPAGATOR Aborts the propagator,
+%   If self is propagating this restores self to before
+%   startpropagator was called. If self is not propagating, this does
+%   nothing.
+end
+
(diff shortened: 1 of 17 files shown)


MATLAB/orbit_estimation/+Orbit/abortpropagator.m
```diff
@@ -0,0 +1,7 @@
+function [self] = abortpropagator(self)
+%ABORTPROPAGATOR Aborts the propagator,
+%   If self is propagating this restores self to before
+%   startpropagator was called. If self is not propagating, this does
+%   nothing.
+end
```
kylekrol: I'm assuming this is used if you want to unapply a deltav or something like that?

kylekrol (commented):
I'd say it looks pretty good! My only large question is if the ability to call geograv a single time is necessary to expose to the user? Or is this because we'll need to spread the orbit estimator calculations over multiple control cycles?

nhz2:
Propagating an orbit that the sat gets from ground could take many control cycles, so that is what `startpropagator`, `abortpropagator`, `finishpropagator`, `numgravcallsleft`, and `onegravcall` handle. `abortpropagator` is a way to stop the propagator, for what ever reason.

kylekrol:
> Propagating an orbit that the sat gets from ground could take many control cycles, so that is what `startpropagator`, `abortpropagator`, `finishpropagator`, `numgravcallsleft`, and `onegravcall` handle. `abortpropagator` is a way to stop the propagator, for what ever reason.

Ah okay yep I'm sorry. I did forget about the fact that we'd need to propegate uplinked orbits. Carry on then!

PR #175, opened 2020-04-07
Added Vector Based DCM Generation Function

### LINKED ISSUE #163: Improve NaN Support in CXX GNC Utility Functions

kylekrol:
# Vector Based DCM Generation Function

Closes #163.

### Summary of changes
- Added a function to generate a right handed DCM given two vectors as a starting basis.
- Cholesky factorization and forward substitution are now in lin as well.
- Fixed a precondition check in the triad function.
- Added a triad NaNs output testcase.

### Ptest Effects
NA

### Testing
See added unit test based on a DCM calculated in MATLAB and the new NaN output testcases.

### Constants
NA

### Documentation Evidence
See inline documentation.

### DIFF
--- a/include/gnc/utilities.hpp
+++ b/include/gnc/utilities.hpp
@@ -77,8 +77,8 @@ constexpr void rotate_frame(lin::Vector<T, 4> const &q, lin::Vector<T, 3> &v);
  *  frame. The y vector's projection along x is removed resulting in the second
  *  basis vector for the new frame. The third is generated assuming a right hand
  *  coordinate system.
- *  There is no explicit handling of NaNs built into this functions; however, a
- *  finite input will always yield a finite result. */
+ *  If the vector x and y are very near to being parallel/anti-parallel, DCM is
+ *  set to NaNs. */
 template <typename T>
 constexpr void dcm(lin::Matrix<T, 3, 3> &DCM, lin::Vector<T, 3> const &x,
     lin::Vector<T, 3> const &y);
--- a/include/gnc/utilities.hpp
+++ b/include/gnc/utilities.hpp
@@ -68,7 +68,7 @@ template <typename T>
 constexpr void rotate_frame(lin::Vector<T, 4> const &q, lin::Vector<T, 3> &v);
 
 /** @fn dcm
- *  @param[out] DCM Direction cosine matrix.
+ *  @param[out] DCM Direction cosine matrix.oth x and y must not be the zero vector.
  *  @param[in]  x
  *  @param[in]  y
  *  Generates a DCM from two reference vector in a particular frame. The DCM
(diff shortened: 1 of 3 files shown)


include/gnc/utilities.hpp
```diff
@@ -67,15 +67,32 @@ constexpr void rotate_frame(lin::Vector<T, 4> const &q,
...
+ *  @param[out] DCM Direction cosine matrix.
+ *  @param[in]  x
+ *  @param[in]  y
+ *  Generates a DCM from two reference vector in a particular frame. The DCM
+ *  will convert from the frame x and y are specified in to the new frame.
+ *  The x input vector is normalized to give the first basis vector of the new
+ *  frame. The y vector's projection along x is removed resulting in the second
+ *  basis vector for the new frame. The third is generated assuming a right hand
+ *  coordinate system.
+ *  There is no explicit handling of NaNs built into this functions; however, a
```
nhz2: Add some preconditions that x and y aren't parallel or anti parallel and are non zero.

nhz2 (approved):

include/gnc/utilities.hpp
kylekrol: Actually, I just added a tolerance check on the vectors being parallel/anti-parallel and set `DCM` to `NaN`s if the tolerance is violated.

include/gnc/utilities.hpp
```diff
@@ -67,15 +67,32 @@ constexpr void rotate_frame(lin::Vector<T, 4> const &q,
...
+ *  @param[out] DCM Direction cosine matrix.
+ *  @param[in]  x
+ *  @param[in]  y
+ *  Generates a DCM from two reference vector in a particular frame. The DCM
+ *  will convert from the frame x and y are specified in to the new frame.
+ *  The x input vector is normalized to give the first basis vector of the new
+ *  frame. The y vector's projection along x is removed resulting in the second
+ *  basis vector for the new frame. The third is generated assuming a right hand
+ *  coordinate system.
+ *  If the vector x and y are very near to being parallel/anti-parallel, DCM is
```
nhz2: Add precondition that x or y aren't zero. Specify what very near means.

nhz2 (changes requested):

test/test_all/utilities_test.cpp
```diff
@@ -37,6 +40,19 @@ void test_utilities_rotate_frame() {
...
 }
 
+void test_dcm() {
+  lin::Vector3f x {1.0f, 1.0f, 0.0f};
+  lin::Vector3f y {1.0f, 0.0f, 1.0f};
+  lin::Matrix3x3f DCM;
+  gnc::utl::dcm(DCM, x, y);
+  lin::Matrix3x3f ans {
+    0.7071f,  0.7071f,  0.0f,
+    0.4082f, -0.4082f,  0.8165f,
```
nhz2, line 51: Add a test for NAN output case.

kylekrol:
This has sort of expanded beyond it's original scope; I ended up adding a triad NaNs testcase because that was the only other GNC function missing. Ended up being missing absolute value call in triad as well.

nhz2 (approved):

PR #176, opened 2020-04-12
C++ orbit class

nhz2:
# Orbit class

### Interface with flight software
I plan on using Orbit class as a state field in the orbit estimation control tasks. Also the best estimate of the satellites current orbit will have to be downlinked and sent to the other satellite. This hasn't been fully implemented yet, currently the Orbit class just has a position velocity and time, and can be propagated for short updates. Right now it isn't used in the rest of flight software.

### Summary of changes
The goal is to create a class to handle most of the logic related to storing, serializing, propagating, and checking validity of an orbit. This class will be used by the next version of the orbit estimator. Another goal is to specify, control, and minimize the number of calls to the gravity function, because one call could take about 2ms on a teensy.

- Added an Orbit class
- Added `-Wno-missing-braces` on native so I can compile on my mac.

The next updates to Orbit will add: 
- More validity checks. #372 
- Long Orbit propagator that can run "asynchronously" (aka take more than one control cycle to finish). #373
- Python or Matlab wrapper.
- Bit compression.

### Testing
I have unit testing in `test/test_orbit/test_orbit.cpp`

### Constants
I add some Orbit related constants.

### Documentation Evidence


test/custom_assertions.hpp
```diff
@@ -0,0 +1,83 @@
+#ifndef CUSTOM_ASSERTIONS_HPP_
```
tanishqaggarwal, line 1: If you put this file here, there might be problems with "duplicate definition" errors once the psim pointer on FlightSoftware is updated. We'll see if it happens and if it does, we can hot-fix the issue

test/test_orbit/test_orbit.cpp
```diff
@@ -0,0 +1,372 @@
...
+    energy_end= y.specificenergy(earth_rate_ecef);
+    TEST_ASSERT_FLOAT_WITHIN(1E-2,0.0,(energy_start-energy_mid));
+    TEST_ASSERT_FLOAT_WITHIN(1E-2,0.0,(energy_end-energy_mid));
+}
+
+/**
+ * Test invalid orbit update gives NAN outputs
+ */
+void test_shortupdate_g() {
+    //test invalid orbit update gives NAN outputs
```
tanishqaggarwal: You have a duplicate comment here

tanishqaggarwal (approved):

include/orb/Orbit.h
```diff
@@ -0,0 +1,391 @@
...
+
+namespace orb
+{
+
+GNC_TRACKED_CONSTANT(constexpr static int, PANGRAVORDER,40);
+GNC_TRACKED_CONSTANT(constexpr geograv::Coeff<PANGRAVORDER>, PANGRAVITYMODEL, static_cast<geograv::Coeff<PANGRAVORDER>>(GGM05S));
+
+GNC_TRACKED_CONSTANT(const double, MAXORBITRADIUS, 6378.0E3L+1000.0E3);
+
+GNC_TRACKED_CONSTANT(const double, MINORBITRADIUS, 6378.0E3L);
```
kylekrol, line 54: Do these constants need to be static as well? I think it's a little weird to have something declared like a global in a header file where it may get included in multiple translation units.

include/orb/JacobianHelpers/jacobian_autocoded.h
```diff
@@ -0,0 +1,171 @@
...
+
+/**
+ * \file jacobian_autocoded.hpp
+ * \author Nathan Zimmerberg
+ * \date 10 APR 2020
+ * \brief Helper function to get the jacobian of in Orbit::shortupdate().
+ * \details  Partially auto code from sympy in jacobian_autocoder.py
+ */
+
+#pragma once
```
kylekrol, line 33: I think we should still stay away from GNU extensions even if they're pretty widely supported. If we could stick to `#ifdef` guards even if it's just for the sake of consistency that'd be great.

test/custom_assertions.hpp
kylekrol: It should be totally fine as long as we only use this in tests themselves.

test/test_orbit/test_orbit.cpp
```diff
@@ -0,0 +1,372 @@
...
+
+#define NANOSECONDS_IN_WEEK (7ULL*24ULL*60ULL*60ULL*1'000'000'000ULL)
+
+//0 or 1 that can't be calculated at compile time.
+#ifdef DESKTOP
+#define notcompiletime (rand()%2)
+#else
+#include <Arduino.h>
+#define notcompiletime (millis()%2)
+#endif
```
kylekrol, line 20: Is this just to prevent the compiler from doing everything at compile time?

test/test_orbit/test_orbit.cpp
```diff
@@ -0,0 +1,372 @@
...
+#include <lin.hpp>
+#include <gnc/constants.hpp>
+#include <gnc/config.hpp>
+
+
+//UTILITY MACROS
+#include <unity.h>
+#include "../custom_assertions.hpp"
+
+#define NANOSECONDS_IN_WEEK (7ULL*24ULL*60ULL*60ULL*1'000'000'000ULL)
```
kylekrol: We could, and probably should, just add this to the constants generator for gnc. It's most likely going to be used in multiple places both in gnc and flight software.

test/test_orbit/test_orbit.cpp
```diff
@@ -0,0 +1,372 @@
...
+
+    //test invalid orbit
+    lin::Vector3d r2 {0.0L, 0.0L, 0.0L};
+    lin::Vector3d v2 {941.0211143841228L, 85.66662333729801L, 7552.870253470936L};
+    uint64_t t2= uint64_t(gnc::constant::init_gps_week_number)*NANOSECONDS_IN_WEEK;
+    orb::Orbit y2(t2,r2,v2);
+    TEST_ASSERT_FALSE(y2.valid());
+
+    //test NAN invalid orbit
+    lin::Vector3d r3 {0.0, std::numeric_limits<double>::quiet_NaN(), 0.0};
```
kylekrol: Just an FYI, there are two constants for NaN values: `constant::nan` for doubles and `constant::nan_f` for floats. Using those would be a little cleaner.

include/orb/JacobianHelpers/jacobian_autocoder.py
```diff
@@ -0,0 +1,62 @@
...
+cseofmatrix=sp.cse(expressions)
+
+def csetupletocline(t,real_type):
+    """Returns a string line of c code of t, a tuple from sympy cse.
+        Uses real_type a string of either "double" or "float" """
+    expand_opt = create_expand_pow_optimization(3)
+    return real_type + " " + sp.ccode(Assignment(t[0],expand_opt(t[1])))
+
+[print(csetupletocline(t,"double")) for t in cseofmatrix[0]]
+[print("jac(%d,%d)= %s;"%(i//6,i%6,str(cseofmatrix[1][i]))) for i in range(len(cseofmatrix[1]))]
```
kylekrol, line 62: This is very cool. I didn't know `sympy` had autocoding built in.

include/orb/Orbit.h
```diff
@@ -0,0 +1,391 @@
...
+        //note if position is NAN, these checks will be false.
+        if (r2<MAXORBITRADIUS*MAXORBITRADIUS && r2>MINORBITRADIUS*MINORBITRADIUS){
+            _valid= true;
+        } else {
+            _valid= false;
+            _recef= lin::nans<lin::Vector3d>();
+            _vecef= lin::nans<lin::Vector3d>();
+            _ns_gps_time= 0;
+        }
+    }
```
kylekrol, line 137: This totally works and is. Just as an aside for future reference, if you wanted to ensure all elements of a `lin` object are finite you can use `lin::all(lin::isfinite(v))`.

include/orb/Orbit.h
```diff
@@ -0,0 +1,391 @@
...
+        geograv::Vector in;
+        geograv::Vector g;
+        in.x= r_ecef(0);
+        in.y= r_ecef(1);
+        in.z= r_ecef(2);
+        potential=geograv::GeoGrav(in, g,PANGRAVITYMODEL,true);
+        g_ecef(0)= g.x;
+        g_ecef(1)= g.y;
+        g_ecef(2)= g.z;
+    }
```
kylekrol, line 156: Do we want to put this function in the environment models potentially? It's fairly likely we may it outside of just the orbit propagator. Or externally would we use a far less accurate gravity model?

test/test_orbit/test_orbit.cpp
```diff
@@ -0,0 +1,372 @@
...
+const orb::Orbit gracestart(uint64_t(gnc::constant::init_gps_week_number)*NANOSECONDS_IN_WEEK,{-6522019.833240811L, 2067829.846415895L, 776905.9724453629L},{941.0211143841228L, 85.66662333729801L, 7552.870253470936L});
+
+//grace orbit 100 seconds later
+const orb::Orbit grace100s(uint64_t(gnc::constant::init_gps_week_number)*NANOSECONDS_IN_WEEK+100'000'000'000ULL,{-6388456.55330517L, 2062929.296577276L, 1525892.564091281L},{1726.923087560988L, -185.5049475128178L, 7411.544615026139L});
+
+//grace orbit 13700 seconds later
+const orb::Orbit grace13700s(uint64_t(gnc::constant::init_gps_week_number)*NANOSECONDS_IN_WEEK+13700'000'000'000ULL,{1579190.147083268L, -6066459.613667888L, 2785708.976728437L},{480.8261476296949L, -3083.977113497177L, -6969.470748202005L});
+
+//earth rate in ecef (rad/s)
+lin::Vector3d earth_rate_ecef= {0.000000707063506E-4,-0.000001060595259E-4,0.729211585530000E-4};
```
kylekrol, line 71: Is this something we'd want as a constant globally as well?

platformio.ini
```diff
@@ -6,14 +6,15 @@ extra_scripts =
   pre:tools/constants_generator.py
```
kylekrol, line 6: Do we want to add the Jacobean generator as a prebuild script as well?

kylekrol (changes requested):
By and large looks great. Other than the things I commented on directly in the code, we should probably consider putting the orbit class in the `gnc` namespace as well. It'd make it easier to use some of the existing `gnc` constants and such.
Also, I think we should look into moving the orbit class implementation into it's own translation unit (don't make it header only). Some of the getters and setters are fine to be inlined but for some of the larger functions I think we'll want to grantee they're compiled separately and linked in.

kylekrol:
Oh, and just curious what the `-Wno-missing-braces` was about? Like what syntax was the compiler choking on.

include/orb/Orbit.h
nhz2: I think in C++ const at namespace scope is implicitly static, unless you use extern or something like that. Also the LTO optimizations will remove redundant copies if it gets compiled into multiple translation units.

nhz2:
> Also, I think we should look into moving the orbit class implementation into it's own translation unit (don't make it header only). Some of the getters and setters are fine to be inlined but for some of the larger functions I think we'll want to grantee they're compiled separately and linked in.

That seems like a premature optimization. I don't want to mess around with compiler optimizations options right now, but if you find something that significantly increases teensy run time performance and the tests still pass, let me know.

nhz2:
> Oh, and just curious what the `-Wno-missing-braces` was about? Like what syntax was the compiler choking on.

In FlightSoftware my compiler was warning about something to do with initializing using curly braces and a constructor with multiple inputs vs using an initializer list and some ambiguity where it wanted extra braces around some of the initializer lists.

test/test_orbit/test_orbit.cpp
nhz2: Yeah

stewartaslan (commented):
I have nothing to add at this point beyond what Kyle and Tanishq already said. Overall things look good!

kylekrol:
> > Also, I think we should look into moving the orbit class implementation into it's own translation unit (don't make it header only). Some of the getters and setters are fine to be inlined but for some of the larger functions I think we'll want to grantee they're compiled separately and linked in.
> 
> That seems like a premature optimization. I don't want to mess around with compiler optimizations options right now, but if you find something that significantly increases teensy run time performance and the tests still pass, let me know.

It's not really a runtime concern, more a binary size concern. The is no guarantee that the compiler will recognize the class is used across multiple translation units and compile a single version of the functions that are shared across multiple translation units.

include/orb/Orbit.h
kylekrol: Ah yep right. `const` ensures internal linkage in c++.

include/orb/JacobianHelpers/jacobian_autocoded.h
nhz2: pragma once is used at least 149 times in FlightSoftware. If there ever is a problem its not too hard to fix.

include/orb/JacobianHelpers/jacobian_autocoder.py
nhz2: Yeah, but it only kind of works, I had to manually edit the output before actually using it. It has a hard time with powers and loves to use pow, instead of a combo of sqrt and multiplication. The coolest thing I think is that `sympy` understands coordinate systems and vector algebra, so you can write out math without a frame and then get the output in whatever frame at the end. Like this 
```
ECEF0 = CoordSys3D('ECEF0')
r= x0*ECEF0.i+y0*ECEF0.j+z0*ECEF0.k
v_E= vx0*ECEF0.i+vy0*ECEF0.j+vz0*ECEF0.k
earth_axis= ECEF0.k
earth_rate= w*earth_axis

v_I= earth_rate.cross(r)+v_E

#inertial frames
ECEFhalf = ECEF0.orient_new_axis('ECEFhalf', w*dt/2,earth_axis)
ECEF1 = ECEF0.orient_new_axis('ECEF1', w*dt,earth_axis)
```

And then to get the output in ECEF1:
`r_ecef= r.to_matrix(ECEF1)`

include/orb/Orbit.h
nhz2: Yeah, I am trying to only use this function in the orbit propagator because it is so expensive, but it is static so any other function can call it if needed.

test/test_orbit/test_orbit.cpp
nhz2: Maybe? This is just a test value, the real earth angular rate changes over time, I'm still not sure how to best deal with that. I'm leaving it exposed as an input for now.

kylekrol: Gothca - makes sense. If it changes very slowly we could leave it as an extern constant later on and then update it from the ground.

platformio.ini
nhz2: No, because the output needs to be manually edited and checked before its used.

include/orb/JacobianHelpers/jacobian_autocoded.h
kylekrol: Pretty much every use of pragma once in fsw comes from the Arduino Json library and not our own code, but I'm not gonna beat a dead horse.

platformio.ini
kylekrol: Yep, wasn't aware there was a manual step to it as well.

kylekrol (approved):

PR #179, opened 2020-04-16
Multi cycle orbit propagator

nhz2:
# C++ Orbit class Multi Cycle Propagation

The main goal is to split up the task of propagating Orbits for a long time between multiple control cycles. If an Orbit gets send from ground and needs to be updated to the current time this will ensure the Orbit gets updated over many control cycles, so the flight computer won't stall.

### Summary of changes
- Added a multi cycle propagator to the Orbit class with an interface of:
     1. `void startpropagating(const uint64_t& end_gps_time_ns, const lin::Vector3d& earth_rate_ecef)`
     2. `int numgravcallsleft() const`
     3.  `void onegravcall()`
     4. `void finishpropagating()`

### Ptest Effects
None

### Testing
Unit testing in `test_orbit.cpp`.

### Constants
```
/** The maximum size of a 7 grav call higher order step (ns). */
GNC_TRACKED_CONSTANT(static const int64_t,maxlongtimestep,100'000'000'000LL);
/** The maximum size of a 1 grav call step (ns). */
GNC_TRACKED_CONSTANT(static const int64_t,maxshorttimestep,200'000'000LL);
```
These constants are internal to the Orbit class propagator implementation, which is tested in `test_orbit.cpp`. 

### Documentation Evidence


include/orb/Orbit.h
```diff
@@ -325,8 +330,8 @@ class Orbit {
         r_ecef0= rel_r+ orb_r;
         v_ecef0= rel_v+ lin::cross(omega,orb_r);
         // step 5b rotate back to ecef
-        _recef= dcm_ecefhalf_ecef0*dcm_ecefhalf_ecef0*r_ecef0;
-        _vecef= dcm_ecefhalf_ecef0*dcm_ecefhalf_ecef0*v_ecef0;
+        _recef= dcm_ecefhalf_ecef0*(dcm_ecefhalf_ecef0*r_ecef0);
+        _vecef= dcm_ecefhalf_ecef0*(dcm_ecefhalf_ecef0*v_ecef0);
```
kylekrol: Good efficient catch! Although, it'd be better if you forced a temporary to be created after the first matrix multiplication. Like:

    _recef= dcm_ecefhalf_ecef0*(dcm_ecefhalf_ecef0*r_ecef0).eval();

and similarly for `_vecef`. Without the eval call, each component of `dcm_ecefhalf_ecef0*r_ecef0` is calculated again every time it's required for the second matrix-vector multiplication.

include/orb/Orbit.h
```diff
@@ -388,5 +393,261 @@ class Orbit {
...
+        _recef= (dcm_ecef_ecef0*_recef).eval();
+        _vecef= (dcm_ecef_ecef0*_vecef).eval();
+        // remove cross r term from velocity
+        _vecef= _vecef-lin::cross(_earth_rate_ecef,_recef);
+    }
+
+
+
+
+
```
kylekrol: Is this extra white-space intentional?

include/orb/Orbit.h
```diff
@@ -388,5 +393,261 @@ class Orbit {
...
+    /** higher order step total dt (s).*/
+    double _currentdt{0};
+    /** mu/(a*a*a) where a is the semimajor axis of the referance orbit (MKS units). */
+    double _mu_a3;
+    /** x axis of reference orbit (m).*/
+    lin::Vector3d _x;
+    /** y axis of reference orbit (m).*/
+    lin::Vector3d _y;
+    /** reference orbit rate (rad/s).*/
+    lin::Vector3d _omega;
```
kylekrol, line 474: This x and y axis of the reference orbit is describing a "close" circular orbit with identical plane and semi-major axis? `_x` stores the positional information and `_y` is set along the velocity vector but still has an identical magnitude to `_x`?

include/orb/Orbit.h
```diff
@@ -388,5 +393,261 @@ class Orbit {
...
+            //convert to relative ecef0
+            _relativize_helper();
+            int64_t deltatime= _targetgpstime-_ns_gps_time;
+            int signofdt= (deltatime<0)?-1:1;
+            int64_t currentdtns;
+            if (_numgravcallsleft>=7){
+                // do a long step
+                if (std::abs(deltatime)>=std::abs(maxlongtimestep)){
+                    // take a full time step
+                    currentdtns= signofdt*maxlongtimestep;
```
kylekrol, line 601: Again, just to make sure I'm understanding this correctly, this is a 7th(ish) order fixed step size orbit propagator correct?
_Not sure about seventh order but wanna make sure it's fixed step size!_

include/orb/Orbit.h
```diff
@@ -388,5 +393,261 @@ class Orbit {
...
+        lin::Vector3d rel_v= _vecef;
+        // drift
+        rel_r= rel_r+rel_v*dt*0.5;
+        // step 3a calc acceleration at the half step
+        double halft= _t+0.5*dt;
+        lin::Vector3d g_ecef0= _get_rel_accel_helper(halft,rel_r);
+        // step 3b kick velocity
+        rel_v= rel_v + g_ecef0*dt;
+        // step 4 drift
+        rel_r= rel_r+rel_v*dt*0.5;
```
kylekrol, line 639: ~This factor of `0.5` seems slightly odd here...? I don't know enough obviously to know for sure and it very well may belong though lmao~
Not gonna argue with the grace data test haha

kylekrol (approved):
All looks pretty great to me and seems to work well given the tests. Just the one `lin` optimization and a few comments/questions.

include/orb/Orbit.h
nhz2: Yeah, kind of. Right now the equation is 
```
double theta= _t*lin::norm(_omega);
double costheta= std::cos(theta);
double sintheta= std::sin(theta);
lin::Vector3d orb_r= _x*costheta+_y*sintheta;
```
But there is a lot of redundant info here.

include/orb/Orbit.h
nhz2: The higher order propagator step right now works like this, first it converts position and velocity in ecef to relative inertial coordinates to a close reference circular orbit. Then it does a series of drift-kick-drift steps (see https://en.wikipedia.org/wiki/Leapfrog_integration ) where a drift is `rel_r= rel_r+rel_v*dt*0.5;` and a kick is `rel_v= rel_v + g_ecef0*dt;` For the low order step(2nd ish) there is just one drift-kick-drift, for the higher order step(6th ish) Yoshida coefficients are used to do 7 drift-kick-drifts with a series of d*dt:
```
static const std::array<double,7> d{{0.784513610477560L,
                                        0.235573213359357L,
                                        -1.177679984178870L,
                                        1.315186320683906L,
                                        -1.177679984178870L,
                                        0.235573213359357L,
                                        0.784513610477560L}};
```
Where somehow this magical series of time steps cause some errors to cancel out. Finally when the step(s) are done the relative position and velocity are converted back to ecef.

kylekrol: Yeah no makes sense. But it's fixed step in the sense that there is no adaptive step size mechanism like `ode45` or `ode23`. Not at all saying that I think we need it, just trying to confirm as observation.

include/orb/Orbit.h
kylekrol: Ahh okay that makes more sense! Thanks.

PR #182, opened 2020-04-24
Ground Propagator

### LINKED ISSUE #181: Add a C++ Ground Propagator Class
The orbit estimator needs to be able to propagate orbits sent from ground to the current time without too much computation a control cycle.

nhz2:
# Ground Propagator

Fixes #181 

### Summary of changes
- Added a GroundPropagator class to propagate orbits sent from ground.
- Added to docs of `Orbit::numgravcallsleft()` that it returns 0 if the Orbit is invalid.

### Ptest Effects
None

### Testing
Unit tests are in `test_groundpropagator.cpp`
Let me know if there is some edge case I'm missing that might break the estimator.

Under normal conditions, this estimator just propagates the most recently uplinked Orbit.

The main goal of the testing is to ensure there is no input that will break the estimator so it can't recover if ground does eventually uplink a good Orbit, and the sat is given enough time to propagate it to the current time.

Also the testing should ensure that too many Orbits getting uplinked won't overload the estimator and prevent it from making progress.

### Documentation Evidence

### DIFF
--- a/include/orb/GroundPropagator.h
+++ b/include/orb/GroundPropagator.h
@@ -131,9 +131,9 @@ class GroundPropagator {
      */
     void one_grav_call(){
         if(current.numgravcallsleft()){
-            current.one_grav_call();
+            current.onegravcall();
         } else { 
-            catching_up.one_grav_call();
+            catching_up.onegravcall();
         }
         //sort stuff
         if (catching_up.valid() && catching_up.numgravcallsleft()<=current.numgravcallsleft()){
--- a/include/orb/GroundPropagator.h
+++ b/include/orb/GroundPropagator.h
@@ -127,13 +127,13 @@ class GroundPropagator {
     /** Do one grav call to move the propagator towards the current time if needed.
      * The Orbit closest to finish propagating is propagated first.
      * 
-     * grav calls: 1 or 0 if totalnumgravcallsleft() == 0 
+     * grav calls: 1 or 0 if total_num_grav_calls_left() == 0 
      */
-    void onegravcall(){
+    void one_grav_call(){
         if(current.numgravcallsleft()){
-            current.onegravcall();
+            current.one_grav_call();
         } else { 
-            catching_up.onegravcall();
+            catching_up.one_grav_call();
         }
         //sort stuff
         if (catching_up.valid() && catching_up.numgravcallsleft()<=current.numgravcallsleft()){
@@ -147,7 +147,7 @@ class GroundPropagator {
      *
      * grav calls: 0
      */
-    int totalnumgravcallsleft(){
+    int total_num_grav_calls_left() const{
         return current.numgravcallsleft()+catching_up.numgravcallsleft()+to_catch_up.numgravcallsleft();
     }
 
@@ -155,7 +155,7 @@ class GroundPropagator {
      * 
      * grav calls: 0
      */
-    Orbit bestestimate() const{
+    Orbit best_estimate() const{
         return current;
     }
 
@@ -163,7 +163,7 @@ class GroundPropagator {
      * 
      * grav calls: 0
      */
-    void resetorbits(){
+    void reset_orbits(){
         Orbit invalid;
         current=invalid;
         catching_up=invalid;
--- a/include/orb/GroundPropagator.h
+++ b/include/orb/GroundPropagator.h
@@ -60,7 +60,7 @@ namespace orb
  * Also to_catch_up replaces catching_up if catching_up finishes propagating.
  */
 class GroundPropagator {
-  private:
+  public:
     /** Current Orbit estimate.*/
     Orbit current;
 
@@ -78,7 +78,6 @@ class GroundPropagator {
      */
     Orbit to_catch_up;
 
-  public:
     /**
      * Construct GroundPropagator.
      *
@@ -94,25 +93,11 @@ class GroundPropagator {
      * @param[in] earth_rate_ecef: The earth's angular rate in ecef frame ignored for orbits already propagating(rad/s).
      */
     void input(const Orbit& ground_data, const uint64_t& gps_time_ns, const lin::Vector3d& earth_rate_ecef){
-        //add to queue
-        if (ground_data.valid()){
-            if(!current.valid()){
-                current= ground_data;
-            } else if(!catching_up.valid()){
-                catching_up= ground_data;
-            } else{
-                to_catch_up= ground_data;
-                //Assuming the new ground_data is closer to the current time 
-                //than previous ground data, over writing this shouldn't be losing any important 
-                //information. 
-                //If that assumion is wrong, then the estimator isn't optimal, but it still works.
-            }
-        }
         //start propagators
         current.startpropagating(gps_time_ns,earth_rate_ecef);
         catching_up.startpropagating(gps_time_ns,earth_rate_ecef);
         to_catch_up.startpropagating(gps_time_ns,earth_rate_ecef);
-        //sort stuff
+        //presort stuff
         if (to_catch_up.valid() && to_catch_up.numgravcallsleft()<=catching_up.numgravcallsleft()){
             catching_up= to_catch_up;
             to_catch_up= Orbit();
@@ -121,6 +106,19 @@ class GroundPropagator {
             current= catching_up;
             catching_up= to_catch_up;
             to_catch_up= Orbit();
+        }
+        //add to sorted queue
+        if (ground_data.valid()){
+            to_catch_up= ground_data;
+            to_catch_up.startpropagating(gps_time_ns,earth_rate_ecef);
+            if(!current.valid() || to_catch_up.numgravcallsleft()<=current.numgravcallsleft()){
+                current= to_catch_up;
+                to_catch_up= orb::Orbit();
+                catching_up= orb::Orbit();
+            } else if(!catching_up.valid() || to_catch_up.numgravcallsleft()<=catching_up.numgravcallsleft()){
+                catching_up= to_catch_up;
+                to_catch_up= orb::Orbit();
+            }
         }
     }
 
--- a/include/orb/GroundPropagator.h
+++ b/include/orb/GroundPropagator.h
@@ -38,6 +38,14 @@ namespace orb
 
 /**
  * Class to propagate orbits sent from ground.
+ * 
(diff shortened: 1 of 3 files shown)


include/orb/GroundPropagator.h
```diff
@@ -0,0 +1,173 @@
...
+     * If catching_up is valid then catching_up is more recently input than current.
+     */
+    Orbit catching_up;
+
+    /** Orbit to catch up to current time once catching_up is done catching up.
+     * If to_catch_up is valid then catching_up is valid.
+     * If to_catch_up is valid then catching_up.numgravcallsleft()<to_catch_up.numgravcallsleft().
+     * If to_catch_up is valid then to_catch_up is more recently input than catching_up.
+     */
+    Orbit to_catch_up;
```
tanishqaggarwal, line 79: If these items are frequently traded amongst themselves, can they be references or pointers? Orbit is an extremely heavy object.

include/orb/GroundPropagator.h
```diff
@@ -0,0 +1,173 @@
...
+                //If that assumion is wrong, then the estimator isn't optimal, but it still works.
+            }
+        }
+        //start propagators
+        current.startpropagating(gps_time_ns,earth_rate_ecef);
+        catching_up.startpropagating(gps_time_ns,earth_rate_ecef);
+        to_catch_up.startpropagating(gps_time_ns,earth_rate_ecef);
+        //sort stuff
+        if (to_catch_up.valid() && to_catch_up.numgravcallsleft()<=catching_up.numgravcallsleft()){
+            catching_up= to_catch_up;
```
tanishqaggarwal, line 102: Is there a copy assignment operator defined on Orbit? Might be necessary if Orbit contains pointers

include/orb/GroundPropagator.h
```diff
@@ -0,0 +1,173 @@
...
+            to_catch_up= Orbit();
+        }
+    }
+
+    /** Return the total number of grav calls need to finish propagating all the orbits.
+     *
+     * grav calls: 0
+     */
+    int totalnumgravcallsleft(){
```
tanishqaggarwal: Function can be marked const

include/orb/GroundPropagator.h
```diff
@@ -0,0 +1,173 @@
...
+            to_catch_up= Orbit();
+        }
+    }
+
+    /** Do one grav call to move the propagator towards the current time if needed.
+     * The Orbit closest to finish propagating is propagated first.
+     * 
+     * grav calls: 1 or 0 if totalnumgravcallsleft() == 0 
+     */
+    void onegravcall(){
```
tanishqaggarwal: To conform with functions everywhere else in FSW and GNC, could these functions be snake-case?

test/test_groundpropagator/test_groundpropagator.cpp
```diff
@@ -0,0 +1,295 @@
...
+
+void test_basic_constructors() {
+    orb::GroundPropagator est;
+    orb::Orbit x= est.bestestimate();
+    TEST_ASSERT_FALSE(x.valid());
+    TEST_ASSERT_EQUAL_INT(est.totalnumgravcallsleft(),0);
+}
+
+
+/** Returns real fake orbits sent from ground.*/
```
tanishqaggarwal, line 756: What is a "real fake" orbit?

test/test_groundpropagator/test_groundpropagator.cpp
```diff
@@ -0,0 +1,295 @@
...
+        break;
+    case 112:
+        //return an old orbit
+        r.startpropagating(r.nsgpstime()+1'000'000'000ULL,earth_rate_ecef);
+        r.finishpropagating();
+        return r;
+        break;
+    case 113:
+        //return an old orbit
+        r.startpropagating(r.nsgpstime()+20'000'000'000ULL,earth_rate_ecef);
```
tanishqaggarwal, line 780: How is overflow prevented in nsgpstime()?

test/test_groundpropagator/test_groundpropagator.cpp
```diff
@@ -0,0 +1,295 @@
...
+        return r;
+        break;
+    default:
+        return orb::Orbit();
+        break;
+    }
+}
+
+
+void test_normal_use() {
```
tanishqaggarwal, line 798: The tests after this line are kind of hard to read. Consider adding more spacing and some header comments on each test describing: the purpose of the tests, the test steps, and how the steps demonstrate the purpose.

tanishqaggarwal: Consider breaking out common initializations into a TestFixture class, rather than copy-pasting the initializations into every test.

test/test_groundpropagator/test_groundpropagator.cpp
```diff
@@ -0,0 +1,295 @@
...
+    lastsentorbit.finishpropagating();
+    best= est.bestestimate();
+    TEST_ASSERT_TRUE(best.valid());
+    TEST_ASSERT_FALSE(best.numgravcallsleft());
+    TEST_ASSERT_TRUE(best.nsgpstime()==lastsentorbit.nsgpstime());
+    PAN_TEST_ASSERT_LIN_3VECT_WITHIN(1E-6, best.vecef(), lastsentorbit.vecef());
+    PAN_TEST_ASSERT_LIN_3VECT_WITHIN(1E-3, best.recef(), lastsentorbit.recef());
+}
+
+void test_edge_cases() {
```
tanishqaggarwal: Could you describe what each of these edge cases are, in comments?

tanishqaggarwal (changes requested):
My main concern is with test readability. The propagator interface looks pretty good and readily integrable into FSW.

include/orb/GroundPropagator.h
kylekrol: While this may be a valid concern I feel like the additional complexity in creating some buffer for orbit objects and then passing around pointers to it may not be worth the extra effort. The relative overhead of a call to `memcpy` compared to the actual propagation is so minimal as well.

include/orb/GroundPropagator.h
kylekrol: I don't there are any internal pointers. However, it may be possible to write a more efficient copy assignment operator/constructor if we don't have to copy all of the internal fields into the new object. For example, if we don't need copy assignment while an orbit is propagating and only need to copy over `r`, `v`, and `t` that could save some overhead.

include/orb/GroundPropagator.h
```diff
@@ -0,0 +1,173 @@
...
+        return current.numgravcallsleft()+catching_up.numgravcallsleft()+to_catch_up.numgravcallsleft();
+    }
+
+    /** Returns a reference to the best estimate of the Orbit.
+     * 
+     * grav calls: 0
+     */
+    Orbit bestestimate() const{
+        return current;
+    }
```
kylekrol: How exactly is this going to integrate with the orbit filter, or are we just going to propagate and not filter orbits received from the ground (which would totally be fine)?

kylekrol (commented):
Overall, the propegation logic and everything looks great. I am wondering, however, if we'd want the `bestestimate` function to ever return an orbit that wasn't caught up to the present. I'm not sure if this will ever be useful information for the rest of gnc/fsw/ Thoughts?

test/test_groundpropagator/test_groundpropagator.cpp
nhz2: There is no overflow protection right now, but nsgpstime is a 64 bit unsigned int, so it won't overflow for hundreds of years.

include/orb/GroundPropagator.h
nhz2: Yeah, the main orbit estimator is going to have a ground propagator for both the self and target orbits. These will used to initialize/reset the Kalman filters, and as a backup incase the GPS or CDGPS isn't giving good data and/or the Kalman filter diverges. There will be a state field flag to just use the ground propagators and ignore GPS, which will also be helpful for testing.

include/orb/GroundPropagator.h
nhz2: Yeah, there are no internal pointers.

include/orb/GroundPropagator.h
nhz2: #183 I made an issue to try this in a future PR if needed.

nhz2:
> Overall, the propegation logic and everything looks great. I am wondering, however, if we'd want the `bestestimate` function to ever return an orbit that wasn't caught up to the present. I'm not sure if this will ever be useful information for the rest of gnc/fsw/ Thoughts?

It probably won't be useful for the rest of GNC if the `best_estimate` isn't caught up. Hopefully this will only happen on startup, or if the GPS time and teensy clock get off by a huge amount (like more than 1 second) for some reason. GNC and flight software can check by calling `est.best_estimate().numgravcallsleft()` and `est.best_estimate().valid()` I'm not sure, but hopefully that will just be inlined and won't actually copy the current Orbit into a temporary object and then call numgravcallsleft().

include/orb/GroundPropagator.h
```diff
@@ -0,0 +1,173 @@
...
+        return current.numgravcallsleft()+catching_up.numgravcallsleft()+to_catch_up.numgravcallsleft();
+    }
+
+    /** Returns a reference to the best estimate of the Orbit.
+     * 
+     * grav calls: 0
+     */
+    Orbit best_estimate() const{
+        return current;
+    }
```
kylekrol, line 158: Return an `Orbit const &` here may more explicitly prevent copying (per your comment above). The use also have the the option to generate a copy themselves from the reference like `Orbit orbit = propegator.best_estimate();`

test/test_groundpropagator/test_groundpropagator.cpp
tanishqaggarwal: Ah, I see, nanoseconds in 64 bits is ok.

include/orb/GroundPropagator.h
nhz2: I'm closing the issue because coping will only happen rarely when an Orbit gets caught up or a new Orbit gets uplinked from ground, not every control cycle.

kylekrol (approved):

tanishqaggarwal (approved):
I don't have the time to read through all of the tests this week, but I know they're probably correct and in the interests of moving progress forward, approving this PR

PR #185, opened 2020-04-26
Fix MEX file errors on Arch Linux

stewartaslan:
# Fix MEX file errors on Arch Linux

Fixes #MEX file compilation is carried out successfully and the correct MEX files are generated, however on Arch Linux an error is thrown after the compilation is complete.184

### Summary of changes
- MEX file compilation is carried out successfully and the correct MEX files are generated, however on Arch Linux an error is thrown after the compilation is complete. To fix this, I added try catch statements around each MEX call in generate_mex_code.

### Ptest Effects
None, except that ptest cases requiring MATLAB will now run on Arch Linux.

### Testing
Deleting the MEX files and running main.m produces no errors, and the MEX files are generated.

### Constants
None

### Documentation Evidence
None

### DIFF
--- a/MATLAB/generate_mex_code.m
+++ b/MATLAB/generate_mex_code.m
@@ -5,5 +5,5 @@ geograv_incl = strcat('-I', fullfile(filepath, "../geograv/include"));
 geomag_file = fullfile(filepath, "environmental_models/helper_functions/geomag_wrapper.cpp");
 geograv_file = fullfile(filepath, "environmental_models/helper_functions/geograv_wrapper.cpp");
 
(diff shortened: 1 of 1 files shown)


tanishqaggarwal (approved):

nhz2:
Do all of the tests pass in `run_tests.m`?

stewartaslan:
> Do all of the tests pass in `run_tests.m`?

Running run_tests.m does not return any statements, just says that each test is done. Does this mean the tests passed or is there a results file somewhere that I need to check?

nhz2:
> > Do all of the tests pass in `run_tests.m`?
> 
> Running run_tests.m does not return any statements, just says that each test is done. Does this mean the tests passed or is there a results file somewhere that I need to check?

The tests passed.

Issue #193, opened 2020-04-28
Unit Tests for the GRP Conversion Functions


kylekrol:
We need to add unit tests for the GRP <-> quaternion conversion functions in the gnc utilities.

Issue #191, opened 2020-04-27
Attitude Estimator Numerical Stability


kylekrol:
In short, we've observed some numerical stability issues with the cxx attitude filter implementation. Below are listed some potential trouble spots and any observations/alternate implementations that have been explored.

### Calculation of `x_bar` and `z_bar`

There isn't a ton to say here other than in the initial attitude filter implementation switching between the below methods of calculation `x_bar` and `z_bar` results in very different values when using floats.

The following:

    x_bar = weight_c * state.sigmas[0];
    z_bar = weight_c * state.measures[0];
    for (lin::size_t i = 1; i < 13; i++) {
      x_bar = x_bar + weight_o * state.sigmas[i];
      z_bar = z_bar + weight_o * state.measures[i];
    }

    x_bar = weight_c * state.sigmas[0];
    z_bar = weight_c * state.measures[0];
    for (lin::size_t i = 1; i < 7; i++) {
      x_bar = x_bar + weight_o * (state.sigmas[i] + state.sigmas[i+6]);
      z_bar = z_bar + weight_o * (state.measures[i] + state.measures[i+6]);
    }

I think this stems from the fact that in this particular testcase the sigma points `6` away from each other tend to be the negative of their counterpart. This can result in catastrophic cancellation when all components aren't of a similar magnitude.

### Kalman Gain Calculation

The Kalman gain `K` is defined as being equal to `P_xy * inv(P_vv)` where `inv` denotes the matrix inverse. The straightforward implementation is to calculate `inv(P_vv)` directly as shown below:

    UkfMatrix6x5 K;
    {
      UkfMatrix5x5 Q, R;
      lin::qr(state.P_vv, Q, R);
      lin::backward_sub(R, Q, lin::transpose(Q).eval());
      K = state.P_xy * Q;
    }

However, it's possible to leverage the fact that `P_vv` is a covariance matrix and therefore should be symmetric. It's possible to write `P_vv * transpose(K) = transpose(P_XY)` which then let's us determine `K` as shown below:

    UkfMatrix6x5 K;
    {
      UkfMatrix5x6 M, N;
      UkfMatrix5x5 L = state.P_vv;
      lin::chol(L);
      lin::forward_sub(L, M, lin::transpose(state.P_xy).eval());
      lin::backward_sub(lin::transpose(L).eval(), N, M);
      K = lin::transpose(N);
    }

[...]

Issue #189, opened 2020-04-27
Magnetometer-Only UKF Implementation


kylekrol:
Per the title.

Issue #188, opened 2020-04-27
Create MEX Functions for the Attitude Estimator


kylekrol:
Basically what's described by the title. We need to be able to run the cxx estimator code both in the full MATLAB sim as well as within @stewartaslan's breakout attitude simulation to verify it's correct.

To learn a little about MEX, a good place to look would be the mex files we have in psim already (`MATLAB/environmental_models/helper_functions/*.cpp`) and the MATLAB mex documentation online.

We'll probably need three functions similar to these:

```
function [state] = adcs_estimator_reset_mex(q, gyro_bias, cov)
function [state] = adcs_estimator_triad_reset_mex(t, r_ecef, b_body, s_body)
function [state, q_est, qyro_bias_est, cov_est] = adcs_estimator_update_mex(state,
    t, r_ecef, b_body, s_body, w_body)
```

where `state` is a MATLAB struct that contains information that must persist across calls to the filter. Most notably, this is the attitude estimate quaternion, gyro bias estimate, and state covariance estimate.

The first function, `adcs_estimator_reset_mex`, doesn't need to incorporate any cxx code.

The second, `adcs_estimator_triad_reset_mex`, should call the following function once it's implemented:

```
void attitude_estimator_reset(AttitudeEstimatorState &state,
    double t, lin::Vector3d const &r_ecef, lin::Vector3f const &b_body,
    lin::Vector3f const &s_body);
```

Lastly, `adcs_estimator_update_mex` will need to populate a `gnc::AttitudeEstimatorState` and `gnc::AttitudeEstimatorData` struct to then call:

```
void attitude_estimator_update(AttitudeEstimatorState &state,
    AttitudeEstimatorData const &data, AttitudeEstimate &estimate);
```

PR #194, opened 2020-04-28
Magnetometer & Sun Vector Attitude Filter Implementation

### LINKED ISSUE #190: Magnetometer and Sun Vector UKF Implementation
Per the title.

kylekrol:
# Magnetometer & Sun Vector Attitude Filter Implementation

Fixes #190.

### Summary of changes
- Added GRP attitude representation utility functions to convert to and from utility functions (deferring intensive unit testing to #193).
- Added double and float interfaces for all environment functions (this is helpful for switching between floats and doubles in the filter implementation itself).
- Updated the `q_eci_perifocal` constant to fix the sun vector model; appropriate unit tests were fixed as well.
- Initial attitude filter implementation when we have both magnetometer and sun sensor readings.
- Simple testcase for the attitude filter with magnetometer + sun sensor readings.

On the surface, the attitude filter seems to work. It's difficult to write good test cases using just PIO because numerical precision, at least so far, has been as issue. For example, the Frobenius norm of the update state covariance matrix is only accurate to two significant digits. Also, we currently only have a single testcase.

There are multiple tickets open to track the work on the attitude filter going forward: #187, #188, #189, #191, #193, and #197. This work includes:

* Developing more testcase and looking at long term filter behavior in MATLAB.
* Improving/verifying numerical stability specifically
* Implementing the magnetometer only Kalman gain update step.
* Adding GRP utility function testcases.

### Ptest Effects
NA

### Testing
See the added unit test cases and tickets listed above for further testing improvements to come.

### Constants
| Constant                 | Location             |
|--------------------------|----------------------|
| `gnc::constant::ukf_sigma_v`     | `src/gnc_attitude_estimator.cpp`   |
| `gnc::constant::ukf_sigma_u`     | `sr/gnc_attitude_estimator.cpp`   |
| `gnc::constant::ukf_sigma_b`     | `src/gnc_attitude_estimator.cpp`   |
| `gnc::constant::ukf_sigma_s`     | `src/gnc_attitude_estimator.cpp`   |
| `dt_thresh` (local scope)     | `src/gnc_attitude_estimator.cpp`   |
| `q_factor` (local scope)     | `src/gnc_attitude_estimator.cpp`   |
| `var_q` (local scope)     | `src/gnc_attitude_estimator.cpp`   |
| `var_g` (local scope)     | `src/gnc_attitude_estimator.cpp`   |
| `init_state` (local scope)     | `src/gnc_attitude_estimator.cpp`   |

### Documentation Evidence
See inline documentation and the paper mentioned in the file documentation in `include/gnc/attitude_estimator.hpp`.


src/gnc_environment.cpp
```diff
@@ -68,13 +68,25 @@ void sun_vector(double t, lin::Vector3f &s) {
...
+void magnetic_field(double t, lin::Vector3d const &r, lin::Vector3f &b) {
   geomag::Vector in, out;
   in.x = r(0);
   in.y = r(1);
   in.z = r(2);
   out = geomag::GeoMag(constant::init_dec_year + t / (365.0 * 24.0 * 60.0 * 60.0),in, geomag::WMM2020);
   b = { out.x, out.y, out.z };
 }
+
+void magnetic_field(double t, lin::Vector3d const &r, lin::Vector3d &b) {
```
tanishqaggarwal, line 86: Won't this function cause an infinite recursion?

include/gnc/constants.hpp
```diff
@@ -60,7 +60,7 @@ GNC_TRACKED_CONSTANT(constexpr static double, earth_perihelion_time, -6287982.0)
 
 GNC_TRACKED_CONSTANT(constexpr static double, earth_period, 365.256363004 * 24.0 * 60.0 * 60.0);
 
-GNC_TRACKED_CONSTANT(constexpr static lin::Vector4d, q_eci_perifocal, 0.127456632677880, -0.158124280715206, 0.762378784011859, -0.614434787689722);
+GNC_TRACKED_CONSTANT(constexpr static lin::Vector4d, q_eci_perifocal, 0.127456632677880, -0.158124280715206, 0.762378784011859, 0.614434787689722);
```
nhz2: Why is this changed and does it need to be changed in MATLAB?

test/test_all/environment_test.cpp
```diff
@@ -60,23 +60,23 @@ void test_environment_sun_vector() {
   // Calculate and check for t=0.0 (comparing against MATLAB)
   gnc::env::sun_vector(0.0L, s);
   TEST_ASSERT_FLOAT_VEC_NEAR(1e-5, s, lin::Vector3f({
-     0.997341623364311f,
```
nhz2: Why did this change?

include/gnc/utilities.hpp
```diff
@@ -47,6 +47,27 @@ template <typename T>
 constexpr void quat_cross_mult(lin::Vector<T, 4> const &q1,
         lin::Vector<T, 4> const &q2, lin::Vector<T, 4> &res);
 
+/** @fn quat_to_grp
+ *  @param[in]  q Input quaternion.
+ *  @param[in]  a 
+ *  @param[in]  f
+ *  @param[out] p Generalized rodriquez parameter representation.
```
nhz2:
```suggestion
 *  @param[out] p Generalized Rodrigues parameter representation.
```

include/gnc/utilities.hpp
```diff
@@ -47,6 +47,27 @@ template <typename T>
...
+ *  Converts a quaternion attitude representation to generalized rodriquez
```
nhz2:
```suggestion
 *  Converts a quaternion attitude representation to generalized Rodrigues
```

include/gnc/utilities.hpp
```diff
@@ -47,6 +47,27 @@ template <typename T>
...
+ *  parameters. There is no explicit handling of NaNs built into this function;
+ *  however, a finite input will always yield a finite result.
+ *  Source: "Unscented Filtering for Spacecraft Attitude Estimation"
+ *          by Markley and Crasidis */
+template <typename T>
+constexpr void quat_to_qrp(lin::Vector<T, 4> const &q, T a, T f, lin::Vector<T, 3> &p);
+
+/** @fn grp_to_quat
+ *  @param[in]  p Input rogriquez parameters.
```
nhz2:
```suggestion
 *  @param[in]  p Input Rodrigues parameters.
```

include/gnc/attitude_estimator.hpp
```diff
@@ -144,8 +181,20 @@ struct AttitudeEstimate {
   bool is_valid;
 };
 
-/** @fn attitude_estimator_reset
- *  Initializes the attitude filter state using the triad method.
+/** @brief Initializes the attitude filter state.
+ *
+ *  @param[out] state      Attitude estimator state to be reset/initialized.
+ *  @param[in]  t          Time since the PAN epoch (units of seconds).
+ *  @param[in]  q_body_eci Quaternion rotating from the body frame to ECI.
```
nhz2:
```suggestion
 *  @param[in]  q_body_eci Quaternion rotating from ECI to the body frame.
```

src/gnc_attitude_estimator.cpp
```diff
@@ -55,26 +387,92 @@ AttitudeEstimate::AttitudeEstimate()
...
-  GNC_ASSERT_NORMALIZED(q_eci_body);
+    double t, lin::Vector3d const &r_ecef, lin::Vector3f const &b_body,
+    lin::Vector3f const &s_body) {
+  GNC_ASSERT_NORMALIZED(s_body);
 
   // TODO : Implement this
-  state = AttitudeEstimatorState();
+  // https://github.com/pathfinder-for-autonomous-navigation/psim/issues/187
+  lin::Vector4f q_eci_body = lin::nans<float, 4, 1>();
+  attitude_estimator_reset(state, t, q_eci_body);
```
nhz2:
```suggestion
  lin::Vector4f q_body_eci = lin::nans<float, 4, 1>();
  attitude_estimator_reset(state, t, q_body_eci);
```

src/gnc_attitude_estimator.cpp
```diff
@@ -55,26 +387,92 @@ AttitudeEstimate::AttitudeEstimate()
...
-  state = AttitudeEstimatorState();
-  estimate = AttitudeEstimate();
+  // Ensure we have a valid state and all the required inputs
+  if (!state.is_valid ||
+      !lin::all(lin::isfinite(data.r_ecef)) ||
+      !lin::all(lin::isfinite(data.b_body)) ||
+      !lin::all(lin::isfinite(data.w_body)) ||
+      !lin::isfinite(data.t)) {
+    // Set the state and estimate to be invalid
+    state = AttitudeEstimatorState();
```
nhz2, line 442: Can the filter just propagate without sensor readings? Or does the state need to be reset.

src/gnc_attitude_estimator.cpp
```diff
@@ -28,16 +28,348 @@
...
+    {
+      UkfMatrix5x5 Q, R;
+      lin::qr(state.P_vv, Q, R);
+      lin::backward_sub(R, Q, lin::transpose(Q).eval()); // Q = inv(P_yy)    
+      K = state.P_xy * Q;
+    }
+
+    // Calculate this steps measurement
+    UkfVector5 z_new {
+      lin::atan(data.s_body(1) / data.s_body(0)),
```
nhz2, line 356: This can be NAN if `s_body(0)==0`

include/gnc/constants.hpp
kylekrol: So at some point this got changed in MATLAB but wasn't updated in the C++. As a result, our sun vector environment function was just wrong. Stewart and I figured this out while debugging the filter.

test/test_all/environment_test.cpp
kylekrol: Same reason as above. The sun vector model got updated in MATLAB at this point and not in C++.

src/gnc_attitude_estimator.cpp
kylekrol: You can't just propagate without sensor readings, at least that's how it's coded now. In terms of our use case however, there won't be a cc where we don't have at least magnetometer and gyroscope data. If for some reason we don't have that information, we have bigger problems than our attitude filter too lol

src/gnc_attitude_estimator.cpp
kylekrol: This is a little concerning... However, in cxx a finite float over zero results in plus or minus infinity (IEEE floats at least) and according to the documentation the `std::atan` function (which is what `lin::atan` essentially calls) returns `pi/2` and `-pi/2` for `inf` and `-inf` respectively. In theory we should be alright.
This code will get replaced when the new sun vector variance handling scheme is incorporated (issue for that coming soon).

Links:
https://en.cppreference.com/w/cpp/numeric/math/atan
https://stackoverflow.com/questions/4745311/c-division-by-0

test/test_all/environment_test.cpp
nhz2: In this branch in MATLAB I'm getting:
```
>> config
>> env_sun_vector(0)

ans =

    0.9973
   -0.0668
   -0.0290
```

nhz2: This makes sense because PAN epoch is in march, near the vernal equinox.

src/gnc_environment.cpp
```diff
@@ -68,13 +68,25 @@ void sun_vector(double t, lin::Vector3f &s) {
   utl::rotate_frame(q_eci_perifocal_f, s);
 }
 
-void magnetic_field(double t, lin::Vector3f const &r, lin::Vector3f &b) {
+void sun_vector(double t, lin::Vector3d &s) {
```
tanishqaggarwal, line 71: Same with this one

tanishqaggarwal (approved):
Just a couple of comments

src/gnc_environment.cpp
kylekrol: No, because of the float `lin::Vector3f` overload just above it.

src/gnc_environment.cpp
kylekrol: Same with this one; there is the float vector overload.

test/test_all/environment_test.cpp
kylekrol: Fixed.

include/gnc/constants.hpp
kylekrol: Reverted these changes and fixed the attitude filter testcase accordingly.

nhz2 (approved):

PR #204, opened 2020-05-03
More Orbit Valid Checks

nhz2:
# More Orbit Valid Checks

### Summary of changes
- Added more Orbit validity checks.
I tried to approximately check that the Orbit will stay between `MINORBITRADIUS` and `MAXORBITRADIUS` using specific orbital energy and specific orbital angular momentum. 
- Changed the gps time in nanoseconds from `uint64_t` to `int64_t`

### Testing
Added more unit tests in `test_validity_checks()`

### Constants
I added MAXGPSTIME_NS and MINGPSTIME_NS
which are about 20 years before and after PAN epoch, Orbits are invalid if they have a time outside this range.

### Documentation Evidence

### DIFF
--- a/include/gnc/constants.hpp
+++ b/include/gnc/constants.hpp
@@ -76,7 +76,7 @@ GNC_TRACKED_CONSTANT(constexpr static float, J_wheel, 135.0e-7f);
 
 GNC_TRACKED_CONSTANT(constexpr static float, w_wheel_max, 677.0f);
 
-GNC_TRACKED_CONSTANT(constexpr static uint64_t, NANOSECONDS_IN_WEEK, 7ULL*24ULL*60ULL*60ULL*1'000'000'000ULL);
+GNC_TRACKED_CONSTANT(constexpr static int64_t, NANOSECONDS_IN_WEEK, 7ULL*24ULL*60ULL*60ULL*1'000'000'000ULL);
 
 }  // namespace constant
 }  // namespace gnc
--- a/include/orb/GroundPropagator.h
+++ b/include/orb/GroundPropagator.h
@@ -92,7 +92,7 @@ class GroundPropagator {
      * @param[in] gps_time_ns: Time to propagate to (ns).
      * @param[in] earth_rate_ecef: The earth's angular rate in ecef frame ignored for orbits already propagating(rad/s).
      */
-    void input(const Orbit& ground_data, const uint64_t& gps_time_ns, const lin::Vector3d& earth_rate_ecef){
+    void input(const Orbit& ground_data, const int64_t& gps_time_ns, const lin::Vector3d& earth_rate_ecef){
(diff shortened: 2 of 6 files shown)


srutiv:
shouldn't line 146 of 'Orbit.h' be "energy = ((v_relative^2/2)- (mu/r))" ? right now it looks like "energy = ((v_relative/2)- (mu/sqrt(r))).  Specific orbital angular momentum check looks good.

nhz2:
> shouldn't line 146 of 'Orbit.h' be "energy = ((v_relative^2/2)- (mu/r))" ? right now it looks like "energy = ((v_relative/2)- (mu/sqrt(r))). Specific orbital angular momentum check looks good.

I added some comments: lin::fro is Frobenius (aka Euclidean) norm squared, so velocity should be squared already. `r2` is also r^2 so the square root just makes it r.

srutiv:
> > shouldn't line 146 of 'Orbit.h' be "energy = ((v_relative^2/2)- (mu/r))" ? right now it looks like "energy = ((v_relative/2)- (mu/sqrt(r))). Specific orbital angular momentum check looks good.
> 
> I added some comments: lin::fro is Frobenius (aka Euclidean) norm squared, so velocity should be squared already. `r2` is also r^2 so the square root just makes it r.

Ahh okay. Looks good then

include/gnc/constants.hpp
```diff
@@ -76,7 +76,7 @@ GNC_TRACKED_CONSTANT(constexpr static float, J_wheel, 135.0e-7f);
 
 GNC_TRACKED_CONSTANT(constexpr static float, w_wheel_max, 677.0f);
 
-GNC_TRACKED_CONSTANT(constexpr static uint64_t, NANOSECONDS_IN_WEEK, 7ULL*24ULL*60ULL*60ULL*1'000'000'000ULL);
+GNC_TRACKED_CONSTANT(constexpr static int64_t, NANOSECONDS_IN_WEEK, 7ULL*24ULL*60ULL*60ULL*1'000'000'000ULL);
```
kylekrol, line 79: I'm assuming the shift in data type is because of python's handling of integers? Is this a concern for flight that'd we'd overflow the signed value?

kylekrol (approved):
Just the concern about `uint64_t` versus `int64_t`. Otherwise, it looks good.

include/gnc/constants.hpp
nhz2: I was doing a lot of casting to int64_t for subtractions, and the Matlab python compatibility thing. There shouldn't be an overflow concern because 2^63 ns is about 300 years, I also added `MAXGPSTIME_NS` and `MINGPSTIME_NS`, where Orbits initialized out of that range are invalidated, so hopefully nothing weird can happen.

PR #211, opened 2020-05-09
Orbit Controller Interface

kylekrol:
# Orbit Controller Interface

### Summary of changes
- Added an orbit controller interface.
- Simple construct tests

This PR only includes the orbit controller interface with very limited unit testing. The following tickets will be opened to track future work:

- Orbit controller implementation (@stewartaslan and I have mostly complete on another branch)
- Orbit controller mex functions
- Orbit controller unit testing in CI
- Orbit controller testing in MATLAB

Let me know if it looks alright.

### Ptest Effects
NA

### Testing
See tickets that will be opened above.

### Constants
NA

### Documentation Evidence
Inline docs.


include/gnc/orbit_controller.hpp
```diff
@@ -0,0 +1,139 @@
...
+ *  @ingroup orbit_controller
+ */
+struct OrbitActuation {
+  lin::Vector3d J_eci;        //!< Recommended impulse vector (Ns).
+  float phase_till_next_node; //!< Phase unil next firing (radians).
+  /** @brief Defaults everything's value to NaN. */
+  OrbitActuation();
+};
+
+#ifndef MEX
```
tanishqaggarwal: For simplicity it'd be better if the structure is
```
#ifdef MEX
the mex function
#else
the non mex function
#endif
```

It reduces the amount of negation to have to read through.

tanishqaggarwal (approved):
Looks good

include/gnc/orbit_controller.hpp
```diff
@@ -0,0 +1,139 @@
...
+namespace gnc {
+
+/** @brief Contains the internal state of an orbit controller.
+ * 
+ *  The internal state only includes a timestamp of the last recommended firing in
+ *  seconds since the PAN epoch. This is to avoid accidently suggesting two
+ *  firings due to a single firing node.
+ * 
+ *  The rest of the state struct serves as a calculation buffer to the controller.
+ *  This is to aid with reducing the stack size required to run it.
```
nhz2, line 51: What is the advantage to reducing stack size, just wondering?

include/gnc/orbit_controller.hpp
```diff
@@ -0,0 +1,139 @@
...
+ *  
+ *  @sa OrbitControllerData
+ *  @sa OrbitActuation
+ * 
+ *  @ingroup orbit_controller
+ */
+struct OrbitControllerState {
+  double t_last_firing; //!< Last firing's timestamp since the PAN epoch (s).
```
nhz2: I'd recommend using `int64_t` nanoseconds since GPS epoch for better compatibility with `orb`

include/gnc/orbit_controller.hpp
```diff
@@ -0,0 +1,139 @@
...
+  lin::Vector3d v_ecef;  //!< Velocity in ECEF (m/s).
+  lin::Vector3d dr_ecef; //!< Relative position of the other satellite in ECEF (m).
+  lin::Vector3d dv_ecef; //!< Relative velocity of the other satellite in ECEF (m/s).
+  /** @brief Defaults everything's value to NaN. */
+  OrbitControllerData();
+};
+
+/** @brief Stores outputs of the control_orbit() function.
+ * 
+ *  These outputs included a recommended impulse vector in Newtons seconds in ECI
```
nhz2: Why ECI and not ECEF?

include/gnc/orbit_controller.hpp
```diff
@@ -0,0 +1,139 @@
...
+ *  and the satellites current position. This allows us to predict when a firing
+ *  is about five minutes out.
+ * 
+ *  @sa OrbitControllerState
+ *  @sa OrbitControllerData
+ * 
+ *  @ingroup orbit_controller
+ */
+struct OrbitActuation {
+  lin::Vector3d J_eci;        //!< Recommended impulse vector (Ns).
```
nhz2, line 105: This probably can be a float.

include/gnc/orbit_controller.hpp
```diff
@@ -0,0 +1,139 @@
...
+  float phase_till_next_node; //!< Phase unil next firing (radians).
+  /** @brief Defaults everything's value to NaN. */
+  OrbitActuation();
+};
+
+#ifdef MEX
```
nhz2, line 113: Why is this Compiler flag needed?

include/gnc/orbit_controller.hpp
nhz2: Initialized to 0.

nhz2 (approved):
Just a few questions.

include/gnc/orbit_controller.hpp
kylekrol: It's more a habit held over from desktop and microcontroller programming where you don't want a stack overflow or to have a stack run into other stacks when running multiple threads.
Since we switched to a single threaded flight computer it doesn't really matter anymore because all we need to worry about is the heap/static memory sections from running into the stack.

include/gnc/orbit_controller.hpp
kylekrol: Initialized to zero or to it's min value? My intention with starting it at NaN was it wouldn't prevent any firing from being scheduled because it was too close to the next one.

include/gnc/orbit_controller.hpp
kylekrol: So we can pull in gains from MATLAB - similar to the attitude controller.

include/gnc/orbit_controller.hpp
kylekrol: Just so flight software didn't have to do the conversion. That being said it's like two lines of code to do the conversion.

kylekrol: I'll swap it back.

include/gnc/orbit_controller.hpp
kylekrol: It also will hide that function from FSW when we are integrating this into a control task.

PR #220, opened 2020-05-10
Added Teensy Build of PSim to CI

kylekrol:
# Added Teensy Build of PSim to CI

### Summary of changes
- Added teensy build to CI to prevent build issues from sneaking in.
- Updated lin to fix compilation issue on teensy.

I'm also going to open up a ticket to clean up our build targets here in PSim and remove some of the duplicate macros that we have flying around.

### Ptest Effects
NA

### Testing
This is the testing.

### Constants
NA

### Documentation Evidence
NA

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -30,4 +30,5 @@ jobs:
         
     - name: Teensy build
       run: |
-        python -m platformio run --verbose -e teensy35
+        python -m platformio run --verbose -e teensy35_ci
+        python -m platformio run --verbose -e teensy36_ci
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -21,10 +21,13 @@ jobs:
         python -m pip install --upgrade pip
         python -m pip install -r requirements.txt
         python -m platformio update
-
-    - name: C++ software tests
-      run: |
         python -m platformio platform install native
-        python -m platformio test --verbose -e github
         python -m platformio platform install teensy
+
+    - name: Desktop software tests
+      run: |
+        python -m platformio test --verbose -e github
+        
+    - name: Teensy build
+      run: |
         python -m platformio run --verbose -e teensy35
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -21,8 +21,13 @@ jobs:
...
 
-    - name: CXX software tests
+    - name: Desktop software tests
       run: |
-        python -m platformio platform install native
         python -m platformio test --verbose -e github
+        
+    - name: Teensy build
+      run: |
+        python -m platformio run --verbose -e teensy35
```
nhz2: Maybe add teensy36 as well.

nhz2:
fails teensy36 tests

```
Processing test_all in teensy36 environment
--------------------------------------------------------------------------------------------------------------------------------------
Building...
.pio/build/teensy36/src/targets/dummy_teensy.cpp.o: In function `setup':
dummy_teensy.cpp:(.text.setup+0x0): multiple definition of `setup'
.pio/build/teensy36/test/test_all/test.cpp.o:test.cpp:(.text.setup+0x0): first defined here
/Users/nathan/.platformio/packages/toolchain-gccarmnoneeabi/bin/../lib/gcc/arm-none-eabi/5.4.1/../../../../arm-none-eabi/bin/ld: Disabling relaxation: it will not work with multiple definitions
.pio/build/teensy36/src/targets/dummy_teensy.cpp.o: In function `loop':
dummy_teensy.cpp:(.text.loop+0x0): multiple definition of `loop'
.pio/build/teensy36/test/test_all/test.cpp.o:test.cpp:(.text.loop+0x0): first defined here
collect2: error: ld returned 1 exit status
*** [.pio/build/teensy36/firmware.elf] Error 1
```

.github/workflows/main.yml
kylekrol: I'll go ahead and do that. And I'm going to make a separate compile target for teensy ci builds only to avoid the issue with testing on teensy.

nhz2 (approved):

PR #226, opened 2020-05-11
Python lin and orb Wrappers

### LINKED ISSUE #215: Python wrappers and basic tests of using orb and lin in python.

nhz2:
# Python lin and orb Wrappers

Fixes #215 

### Python Wrappers
There are python wrappers for some C++ types. I am using [pybind11](https://pybind11.readthedocs.io/en/stable/intro.html) which has great documentation and many examples. pybind11 is just a header only library that makes it easy to write python extensions, https://thomasnyberg.com/what_are_extension_modules.html . 

The C++ sources need to be compiled as a shared library to be imported by python. 
I am using [cppimport](https://github.com/tbenthompson/cppimport) to do this automatically on import but it is theoretically possible to compile using cmake, pio, pip ... 

To test the compilation run `python estimatortest/pythonwrapper/test_compilation.py` Which should print a bunch of compiler errors if there is a mistake.

`psim/estimatortest/pwrap.cpp` Is the main module file.

The giant comment at the top is used in a python script that sets the compiler configuration.
`cfg['dependencies']` is a list of filenames that are checked for changes to see if recompiling is needed.
`cfg['sources']` are the sources to compile.
`cfg['include_dirs']` are the include directories.
`cfg['compiler_args']` are the compiler args, I also made a section for platform specific compiler args because my mac needed `'-mmacosx-version-min=10.9'` to compile.

#### orb wrappers
`psim/estimatortest/orb_ext.cpp` Has the `orb` wrappers.

For now `orb::Orbit` as python type `orb_Orbit` is mostly wrapped.

#### lin wrappers
`psim/estimatortest/lin_ext.cpp` Has the `lin` wrappers.

Most of this file is from https://pybind11.readthedocs.io/en/stable/advanced/pycpp/numpy.html examples.

Because `lin` uses templates, each type that you want to use in python must be specifically wrapped. For example, right now in `lin_ext.cpp` I have:
```
    WRAPLINMATRIX("lin_Matrix6x6d",m,double,6,6,6,6);
    WRAPLINVECTOR("lin_Vector3d",m,double,3,3);
```

These create python types named `lin_Matrix6x6d` and `lin_Vector3d` 
for C++ types `lin::Matrix<double, 6, 6, 6, 6>` and  `lin::Vector<double, 3, 3>`.
These are compatible with numpy arrays, so numpy.array(lin_Vector3d([1.0,2.0,3.0])) will work.
The wrapper doesn't wrap all of the `lin` features, so it's usually best to always convert the lin types into numpy arrays if you want to do math with them in python.

If you need other `lin` types in python make sure to wrap them in this file using the macros.

### Testing
I am using pytest to test the wrappers in CI.

### DIFF
--- a/estimatortest/pythonwrapper/test_lin.py
+++ b/estimatortest/pythonwrapper/test_lin.py
@@ -129,8 +129,6 @@ def testmatix():
     assert np.all(np.array(b)==np.array([[1,2],[3,4]],dtype='f8'))
 
 
-
-
 if __name__ == "__main__":
     pwrap= cppimport.imp("pwrap")
     testvectors()
--- a/estimatortest/pythonwrapper/test_lin.py
+++ b/estimatortest/pythonwrapper/test_lin.py
@@ -1,9 +1,11 @@
 import cppimport
 import numpy as np
 import copy
+if __name__ != "__main__":
+    from pythonwrapper import pwrap
+    #This is because pytest is weird with setting the path
 
 def test_lin():
-    from pythonwrapper import pwrap
     testvectors()
     testmatix()
 
--- a/estimatortest/pythonwrapper/test_lin.py
+++ b/estimatortest/pythonwrapper/test_lin.py
@@ -1,0 +1,136 @@
+import cppimport
+import numpy as np
+import copy
+
+def test_lin():
+    from pythonwrapper import pwrap
+    testvectors()
+    testmatix()
+
+def testvectors():
+    # lin_Vectors can be made from regular lists
+    x= pwrap.lin_Vector3f([1.0,2.0,1.0E-9])
+    x= pwrap.lin_Vector3f([1,2.0,1.0E-9])
+    x= pwrap.lin_Vector3d([1.0,2.0,1.0E-9])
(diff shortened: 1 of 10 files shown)


estimatortest/pythonwrapper/test_lin.py
```diff
@@ -0,0 +1,136 @@
...
+    x[0,0]=5
+    assert np.all(np.array(a)==x)
+    assert np.all(np.array(b)==np.array([[1,2],[3,4]],dtype='f8'))
+
+    a= pwrap.lin_Matrix2x2d(np.array([[1,2],[3,4]]))
+    b= pwrap.lin_Matrix2x2d(a) #b is an independent copy of a
+    x= np.array(a,copy=False)
+    x[0,0]=5
+    assert np.all(np.array(a)==x)
+    assert np.all(np.array(b)==np.array([[1,2],[3,4]],dtype='f8'))
```
kylekrol, line 129: I'm assuming that `f4` vs `f8` is just a `float` vs `double` distinction?

kylekrol (approved):
All looks good to me!

PR #231, opened 2020-09-12
Infrastructure for the new PSim

kylekrol:
# Infrastructure for the new PSim

This PR contains most of the C++ infrastructure code and the Bazel build system for the PSim. It's structured as a library of "models" that will later be linked into small PyBind11 packages which will be used to test GNC software and run the simulation with PTest.

To jump right in and run some tests, execute the following commands from the root of the PSim repository:

    python37 -m venv venv
    source venv/bin/activate
    pip install -r requirements.txt
    bazel test //:test/psim:all

This will compile the GNC library, PSim library, and link both of those into the current set of PSim unit tests written with GoogleTest. The Python virtual environment is required for the model interface autocoder (more on this later).

To get intellisense working with the Bazel build system, take a look at the `tools/bazel-compilation-database.sh` file.

## Motivation

The following outlines the new features we're looking for out of a simulation written in C++ and why we're moving away from MATLAB:

 * Work with an architecture that helps prevent negative work. Previously with the MATLAB simulation, rewriting the truth orbit propagator meant writing over the existing code. This had the added implication that the old code was entirely lost and we were rewriting some of the simpler orbit propagators over and over again when this wasn't fully necessary.
 * Better simulation support for multiple use cases. The MATLAB simulation wasn't very modular at all and when working with the simulation, many changes ended up being breaking. For example, without GNC being fully integrated into flight software, we weren't allowed to add sensor models into the simulation because it'd break things downstream. We need a modular design for simulation models to allow them to be easily swapped in and out.
 * Support a wide range of telemetry but don't needlessly calculate information if it's not needed for the sake of better performance. Essentially, this means we'd like to support lazy evaluation which wouldn't be that easy to handle in MATLAB.
 * MATLAB can't be run in CI.

In addition to the above grievances, we're striving for some other features as well:

 * A clean configuration system to provide initial conditions to the sim and handle other constants (sensor noises, et cetera).

## Architecture

As alluded to above, PSim will be written as a collection of "models". These models will interact with the other two main components of PSim when integrated into a full simulation: the configuration and the state field systems.

### Configuration

At a high level, the configuration system is responsible for parsing configuration files (simple key-value text files) into a set of "parameters" that can be read in by models during sim initialization.

See the following files for more details:

    include/psim/core/configuration.hpp
    include/psim/core/parameter*.hpp
    src/psim/core/configuration.cpp

### State Fields

State fields and a simulations state registry are responsible for tracking and providing access to the state of the simulation as it evolves over simulation steps (doesn't necessarily have to be forward in time but it pretty much always is).

State fields can be registered to the simulation state in two fashions: either as a writable field or a plain state field (presented as a read-only field). A writable field can be written from outside the simulation or another model (say a reaction wheel actuator model that's supposed to present a wheel torque to the attitude dynamics model). Fields registered as read-only can only have telemetry pulled from them from outside the simulation or by another model.

There is one kind of read-only field I'd like to draw special attention to and that's the `template <typename T> class StateFieldLazy<T>`. This field holds a `std::function<T ()>` that can be called if/when the state field is accessed. This helps handle the "support a wide range of telemetry" requirement as we can "teach" the sim how to evaluate a bunch of parameters that may only be accessed once in a blue moon.

See the following files for more details:

    include/psim/core/state*.hpp
    src/psim/core/state.cpp

### Models

Models are where the bulk of future simulation development will be happening. They're responsible for adding there state fields to the simulation state, retrieving additional fields from the state they may need, and specifying how these fields relate and should be stepped forward.

As one could imagine, writing all of the "wiring" code over and over again would get a little annoying. To help combat this and standardize how models should interact with the simulation at large, all model interfaces are to be specified in a YAML file which will then be autocoded into a model base class. This base class interface could have multiple implementations in the form of many derived classes.

Last quick thing to note, a model can also be an ordered list of other models. We call this a `ModelList` and it will be helpful in assembling complex simulations into a single model (important for creating a simulation).

See the following files for more details:

    include/psim/core/model*.hpp
    include/psim/truth/*
    src/psim/core/model*.cpp
    src/psim/truth/*
    tools/autocoder.py

### Simulation

The last stop on the architecture overview is the full on simulation! The only thing the simulation class as a whole does is bring together the configuration, model, and state field systems into a functioning simulation.

See the following files for more details:

    include/psim/core/simulation.hpp
    src/psim/core/simulation.cpp


kylekrol:
Description is in progress. I definitely recommend looking at it commit by commit!

include/psim/core/state_field_base.hpp
```diff
@@ -0,0 +1,156 @@
...
+      throw std::runtime_error("Invalid field type in call to 'cast'.");
+
+    return field_ptr;
+  }
+
+  template <typename T>
+  StateField<T> *cast() {
+    auto *field_ptr = dynamic_cast<StateField<T> *>(this);
+    if (!field_ptr)
+      throw std::runtime_error("Invalid field type in call to 'cast'.");
```
tanishqaggarwal, line 86: This error message is basically useless because it doesn't tell you which field caused the error. I know for a fact this causes hellish debugging experiences in...other settings :) . Can we make these errors more descriptive?

include/psim/core/state_field_lazy.hpp
```diff
@@ -0,0 +1,101 @@
...
+   */
+  std::function<T ()> const _function;
+
+  /** @brief Flag specifying if the currently held value is up to date.
+   */
+  bool mutable _evaluated;
+
+  /** @brief Current value (may or may not be valid).
+   */
+  T mutable _value;
```
tanishqaggarwal, line 61: What's the point of making this mutable if only one function ever accesses it and the function is marked const? Better to separate the function into separate calls, right?

include/psim/core/state_field_valued.hpp
```diff
@@ -0,0 +1,105 @@
...
+/** @brief Writable state field implemenation directly backed by a value.
+ *
+ *  @tparam Underlying type.
+ */
+template <typename T>
+class StateFieldValued : public StateFieldWritable<T> {
+ private:
+  /** @brief Underlying value.
+   */
+  T mutable _value;
```
tanishqaggarwal: Same here. Mutable is a bad idea--for such base-level objects I feel like we should be very deliberate about the class contracts, and add getters/setters rather than trying to use clever language features to make things less verbose.

src/psim/core/configuration.cpp
```diff
@@ -0,0 +1,147 @@
...
+
+  return *parameter_ptr;
+}
+
+ParameterBase const *Configuration::get(std::string const &name) const {
+  auto const iter = _parameters.find(name);
+  return (iter == _parameters.end() ? nullptr : iter->second);
+}
+
+Configuration Configuration::make(std::string const &file) {
```
tanishqaggarwal, line 76: Here I would've considered writing a parser using existing frameworks like GNU bison. Thankfully the parsing is simpler enough right now but if we discover later on that we want more options, such as macros to keep duplicating configurations, it will make our lives a million times easier.

src/psim/core/model.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+/** @file psim/core/model.cpp
+ *  @author Kyle Krol
+ */
+
+#include <psim/core/model.hpp>
+
+namespace psim {
+
+void Model::add_fields(State &state) { }
```
tanishqaggarwal, line 33: Why not just make these part of the header file? This way we can get rid of this whole file.

test/psim/BUILD
```diff
@@ -0,0 +1,6 @@
+load("//:tools/psim.bzl", "psim_cc_test")
```
tanishqaggarwal, line 1: Maybe you already did this but I missed it in the ocean of files:
- I think it's very essential that we document all testable and buildable targets somewhere. Definitely should do it as part of this PR.

test/psim/core/counter.hpp
```diff
@@ -0,0 +1,30 @@
...
+#ifndef TEST_PSIM_CORE_COUNTER_HPP_
+#define TEST_PSIM_CORE_COUNTER_HPP_
+
+#include <psim/core/configuration.hpp>
+#include <psim/core/model.hpp>
+#include <psim/core/parameter.hpp>
+#include <psim/core/state_field_valued.hpp>
+#include <psim/core/types.hpp>
+
+/** @brief Simple model that counts an integer forward for testing purposes.
```
tanishqaggarwal, line 14: If this is only used for testing purposes why keep it out in core/?

BUILD
```diff
@@ -1,11 +1,75 @@
...
+    visibility = ["//visibility:public"],
+    deps = ["@lin//:lin"],
+)
+
+# Let bazel know about our virtual environment.
+#
+# Also see additional options passed with .bazelrc to ensure we're using the
+# virtual environment.
+#
+# TODO : This is a little hacky and could be cleaned up with toolchains.
```
tanishqaggarwal, line 28: Tbh not worth it.

BUILD
```diff
@@ -1,11 +1,75 @@
+load("@rules_python//python:defs.bzl", "py_runtime", "py_binary")
+load("@rules_cc//cc:defs.bzl", "cc_library")
+
+load("//:tools/psim.bzl", "psim_autocode")
+
+# Build flight software's gnc code as a library.
+cc_library(
```
tanishqaggarwal, line 7: Who is the consumer of this library? Note that FSW can't really use this stuff because we're still going with PlatformIO, the Bazel transition wasn't too successful.

test/psim/BUILD
tanishqaggarwal: Also should list any "getting started" steps in the README, e.g. installilng Bazel.

tanishqaggarwal (approved):

include/psim/core/state_field_lazy.hpp
kylekrol: This is marked mutable because a lazy state field should always be seen as a constant state field - i.e. it's value cannot be changed outside of a given model. However, when accessing it's "value" it may not have been calculated yet. Hence the need for it to be marked mutable.

include/psim/core/state_field_valued.hpp
kylekrol: This one here should definitely not be mutable though, you're correct!

src/psim/core/configuration.cpp
kylekrol: Agreed - to start of with however I'd just keep it as is. I don't think there's a need to make it more complicated or the time being.

test/psim/BUILD
kylekrol: Ah yes, correct. I did forget to add in the getting started information to the README.

test/psim/core/counter.hpp
kylekrol: Not sure if I understand the question here? It's only for testing core infrastructure so I just left it in `test/psim/core`?

BUILD
kylekrol: Agreed haha But just in case something breaks down the line I wanted to leave this note here.

BUILD
kylekrol: We'll have a Python module build target that will consume the library. That Python module will then be consumed by PTest!

test/psim/core/counter.hpp
tanishqaggarwal: Oh I'm dumb lol. I didn't see the test/ at the beginning.

src/psim/core/model.cpp
kylekrol: I guess I just left this here because I foresaw us adding some debugging state fields and what not to each individual model. Figured I would just lay the groundwork for that here directly so implementations don't end up in the header but I could just get rid of it.

include/psim/core/state_field_base.hpp
kylekrol: Yeah... I wanted to avoid creating more copies of strings for the state field names however - which is why this error message doesn't include a state field name. I've found that using gdb to give a stacktrace is helpful and basically tells you where the error was from but that may not be maintanable.

kylekrol: How do we feel about opening and infrastructure ticket for this and I can take a look once more of the simulation is spun up?

tanishqaggarwal: Discussed offline that punting this work is ok. It's a lot of nontrivial refactoring.

athena255 (approved):
Nice, this is really clean.

include/psim/core/state_field_base.hpp
kylekrol: See #232.

PR #233, opened 2020-09-20
Matlab vs C++ MEX Attitude Estimation

### LINKED ISSUE #187: Triad-Based Attitude Filter Initialization Function
The UKF attitude filter requires an initial "guess" of our current attitude. On orbit, an easy way to accommodate this would be to initialize the filter once we have a GPS estimate, magnetometer readings, and sun vector reading.

The attitude UKF requires an initial "guess" of our attitude equation prior to simply calling it's update function on every control cycle. On orbit, one easy way to handle this is initializing the attitude filter based on a triad estimate once we have a valid GPS estimate, magnetometer reading, and sun vector reading.

Once #192 is merged, this code will need to be added to `src/gnc_attitude_estimator.cpp` per the inline documentation in `include/gnc/attitude_estimator.hpp` given below:

```
/** @fn attitude_estimator_reset
 *  Initializes the attitude filter state using the triad method.
 *
 *  @param[out] state  Attitude estimator state to be reset/initialized.
 *  @param[in]  t      Time since the PAN epoch (units of seconds).
 *  @param[in]  r_ecef Position in ECEF (units of meters).
 *  @param[in]  b_body Magnetic field measured in the body frame (units of Tesla).
 *  @param[in]  s_body Sun vector measured in the body frame (unit vector).
 *
 *  Uses the current time, position, b field, and sun vector to calculate an
 *  attitude estimate to initialize the filter with. The other filter state
 *  variables are set to default values.
 *
 *  After calling, it's recommended to check the `state.is_valid` flag to ensure
 *  the estimator state was succesfully reset. The reset call will fail is
 *  parameters given failed the triad algorithm - i.e the sun and magnetic field
 *  vector were nearly parallel.
 * 
 *  See the `gnc::utl::triad` documentation for more information. */
void attitude_estimator_reset(AttitudeEstimatorState &state,
    double t, lin::Vector3d const &r_ecef, lin::Vector3f const &b_body,
    lin::Vector3f const &s_body);
```

A good reference to get started would be looking at the old estimator code here which performed triad given a sun vector, magnetic field reading, current position, and timestamp. Here is a [link](https://github.com/pathfinder-for-autonomous-navigation/psim/tree/383d0192b12c7ca2ed1fec8b7f097646a6358c14) to an old version of the repo containing the previous estimator's code. And below is a snippet of the triad method being performed:

    #include <gnc/constants.hpp>
    #include <gnc/environment.hpp>
    #include <gnc/utilities.hpp>

    #include <lin/core.hpp>

    // Determine the magnetic field in ECI (it's given in ECEF)
    env::magnetic_field(t, lin::Vector3f(r_ecef), b_eci);  // b_eci = b_ecef
    env::earth_attitude(t, q);                             // q = q_ecef_eci
    utl::quat_conj(q);                                     // q = q_eci_ecef
    utl::rotate_frame(q, b_eci);                           // b_eci = b_eci

    // Ensure the measured and expected magnetic field are large enough
    float thresh = constant::b_noise_floor * constant::b_noise_floor;
    if ((lin::fro(b_eci) < thresh) || (lin::fro(b_body) < thresh))
      return;

    // Determine the expected sun vector
    env::sun_vector(t, s_eci);

    // Ensure the sun vectors are unit vectors
    GNC_ASSERT_NORMALIZED(s_body);
    GNC_ASSERT_NORMALIZED(s_eci);

    // Perform triad
    // Remember the value of q_body_eci will be set to NaNs if triad fails
    utl::triad(s_eci, (b_eci / lin::norm(b_eci)).eval(), s_body,
        (b_body / lin::norm(b_body)).eval(), q_body_eci);

shihaocao:
# Matlab vs C++ MEX Attitude Estimation

Fixes #187, #188, ???. (Optional; sometimes PRs don't have an associated ticket)

### Summary of changes
- Implement `estimator_triad_reset` - #187
- Mex Wrappers for `estimator_reset`, `estimator_triad_reset`, `estimator_update` - #188 
- Eclipse?
- Matlab UKF restructuring?

### Ptest Effects
No effects?

### Testing
Visual unit tested by looking at graphs?

### Constants
No changes to constant

### Documentation Evidence
Post to ReadTheDocs, or
- Argue that your inline documentation is sufficient.
- Create a issue ticket deferring the documentation task.


shihaocao:
@srutiv 

I made a draft PR for this branch since I think it'd be a good place to try and get this MEX stuff in. I'm not sure if you stopped in a merge-able place. Let me know what you think we should do

src/gnc/attitude_estimator.cpp
```diff
@@ -438,10 +439,35 @@ void attitude_estimator_reset(AttitudeEstimatorState &state,
     double t, lin::Vector3d const &r_ecef, lin::Vector3f const &b_body,
     lin::Vector3f const &s_body) {
   GNC_ASSERT_NORMALIZED(s_body);
-
-  // TODO : Implement this
-  // https://github.com/pathfinder-for-autonomous-navigation/psim/issues/187
+
```
shihaocao, line 441: this function was essentially lifted from #187

MATLAB/attitude_filter/helper_functions/conversion.hpp
```diff
@@ -0,0 +1,53 @@
+#include "mex.hpp"
+#include "mexAdapter.hpp"
+
+#include <gnc/attitude_estimator.hpp>
+#include <lin.hpp>
+
+using namespace matlab::data;
+
+template<lin::size_t N, typename T>
```
tanishqaggarwal, line 18: Docstrings would be helpful to explain what these functions exist for to any future reader of this PR.

src/gnc/attitude_estimator.cpp
```diff
@@ -40,6 +40,7 @@
 #include <lin/math.hpp>
 #include <lin/queries.hpp>
 #include <lin/references.hpp>
+//#include <lin/substitutions/forward_substitution.hpp>
```
tanishqaggarwal: Might as well delete this line?

src/gnc/attitude_estimator.cpp
```diff
@@ -438,10 +439,35 @@ void attitude_estimator_reset(AttitudeEstimatorState &state,
...
   GNC_ASSERT_NORMALIZED(s_body);
-
-  // TODO : Implement this
-  // https://github.com/pathfinder-for-autonomous-navigation/psim/issues/187
+  
   lin::Vector4f q_body_eci = lin::nans<float, 4, 1>();
+  
+  lin::Vector3f b_eci{};
+  lin::Vector4f q{};
+  env::magnetic_field(t, lin::Vector3f(r_ecef), b_eci);  // b_eci = b_ecef
```
tanishqaggarwal: Instead of the comments, why not just add variables that are named things like q_eci_ecef, as part of the code? Then the code is self-documenting.

src/gnc/attitude_estimator.cpp
```diff
@@ -438,10 +439,35 @@ void attitude_estimator_reset(AttitudeEstimatorState &state,
...
+  env::earth_attitude(t, q);                             // q = q_ecef_eci
+  utl::quat_conj(q);                                     // q = q_eci_ecef
+  utl::rotate_frame(q, b_eci);                           // b_eci = b_eci
```
tanishqaggarwal: I think the comment should say something like b_eci = b_ecef? unless I'm misunderstanding.

tanishqaggarwal (changes requested):
For "Tests", I think it would be nice to see the plots and explanations of what they mean.

Also, this software definitely will have ptest affects since updating the psim pointer on FSW will cause AttitudeEstimator to pick up this change in implementation. Although I agree this particular PR won't affect fsw since it doesn't update the pointer, might be worth mentioning in the description.

tanishqaggarwal:
As requested I didn't take a look at any of the MATLAB files, let me know if you'd like me to review any of them.

src/gnc/attitude_estimator.cpp
kylekrol: Because then we are growing the stack for no reason. The intention with these comments is to let you know what is in the variable `b_eci` at a given line. By the end of our calculation's we want it to be the magnetic field in an inertial frame.

src/gnc/attitude_estimator.cpp
kylekrol: Here, `b_eci` was previsouly equal to `b_ecef` (see line 447) and `q` is `q_eci_ecef`. Therefore the transformation moves the magnetic field reading from ECEF to EC.

src/gnc/attitude_estimator.cpp
```diff
@@ -438,10 +439,35 @@ void attitude_estimator_reset(AttitudeEstimatorState &state,
...
+  
   lin::Vector4f q_body_eci = lin::nans<float, 4, 1>();
+  
+  lin::Vector3f b_eci{};
+  lin::Vector4f q{};
+  env::magnetic_field(t, lin::Vector3f(r_ecef), b_eci);  // b_eci = b_ecef
+  env::earth_attitude(t, q);                             // q = q_ecef_eci
+  utl::quat_conj(q);                                     // q = q_eci_ecef
+  utl::rotate_frame(q, b_eci);                           // b_eci = b_eci  
+  env::magnetic_field(t, lin::Vector3f(r_ecef), b_eci);
```
kylekrol: Pretty sure there is also a `lin::cast<float>(r_ecef)` for something like this. It doesn't matter in this instance because the function asks for the vector by value and not a stream. However, if working with `lin::` functions it's better to use `lin::cast<float>(r_ecef)`. You may have to throw a `.eval()` after the call to cast too.

src/gnc/attitude_estimator.cpp
```diff
@@ -438,10 +439,35 @@ void attitude_estimator_reset(AttitudeEstimatorState &state,
...
+  
+  // Ensure the measured and expected magnetic field are large enough
+  float thresh = constant::b_noise_floor * constant::b_noise_floor;
```
kylekrol: We should throw this into something like:
```
{
  float thresh = constant::b_noise_floor * constant::b_noise_floor;
  float fro_b_eci = lin::fro(b_eci);
  if (...) return;
}
```
The compiler may see through it but just in case it doesn't we can save some floating point multiplications.

src/gnc/attitude_estimator.cpp
kylekrol: Yeah we can take this out.

kylekrol: Also add an `@author ...` for yourself at the top if you feel so inclined.

kylekrol (changes requested):
Nothing to huge. I'd just clean up the comment at the top of the attitude estimator implementation and a couple other suggestions.

shihaocao:
- Added docstrings to conversion.hpp
- Added scope to noise floor

kylekrol (approved):

src/gnc/attitude_estimator.cpp
tanishqaggarwal: I heavily disagree with trying to save stack space at the cost of making code more readable. We should only do this if stack space starts proving to be a problem.

At the very least, we should modify the variable name from `b_eci` to something else if it's not always the magnetic field vector in ECI.

tanishqaggarwal: Compilers are very good at eliminating unnecessary stack use so I actually think this concern is unwarranted

kylekrol: This doesn't seem like an issue worth holding this PR up especially considering the code _is_ documented with how the local variable is being mutated. I also prefer this method on handling local variables because it prevents polluting the local namespace with a ton of variables that we don't need and is more explicit with the values we are actually looking to calculate - we have code following this style all over the place in GNC land.

kylekrol: Besides the stack deal it helps we work with intellisense better to be quite honest.

kylekrol: Don't really care though we should just merge this in one way or the other.

kylekrol: Shihao is changing this now I think

tanishqaggarwal (approved):
I like the new change a lot better

shihaocao:
Screenshot from 2020-09-27 18-49-39

shihaocao:
Added the more important graphs for documentation

PR #239, opened 2020-09-26
Implemented First Simulation and Early Python Bindings

kylekrol:
# First Simulation and Early Python Bindings

## Improvements to the Build System

There were many updates to the build system in this PR.
 * All `*.bzl` files were centralized under the new `bazel` directory.
 * A repository rule was added to detect the system's Python headers and create an external repository with a symlink to the headers. This allows Bazel to add them to the included path and build pybind11 modules.
 * A C++14 implementation of `std::variant` was added to support dynamic return types from the simulation to Python.
 * More psim specific Bazel commands were created for convenience.

## GNC Updates

I've updated all of the fixed step size integrators to take a void pointer as an argument to the state update function. This will be useful when the models in psim get more complex! Unit tests were updated for this as well.

## PSim Models

Many more truth models were implemented. The primary ones are specified in the new `*.yml` files under `psim/include/truth/`. These form the basic buildings blocks for a truth sim that outputs sensor values!

_Joint attitude and orbit truth model is still under development_

## PSim Simulations

We've now plugged in a 3-DOF single satellite, orbit simulation into Python! The simulations that Python will consume are listed under `include/psim/simulations` and `src/psim/simulations` and the Python wrappers (along with the other Python code to support simulations) is under `python/psim/**`.

## DIY

To test this out yourself run the following from the psim repository (follow install instructions in the README first):

    bazel run //python:single_orbit

You should have some pretty graphs! I tested this out myself on Mac and Arch.

Screen Shot 2020-09-26 at 8 55 18 PM


.bazelrc
```diff
@@ -1,2 +1,2 @@
-build --cxxopt=-std=c++14
+build --cxxopt=-std=c++14 --cxxopt=-O3 --cxxopt=-Werror --cxxopt=-Wall
```
kylekrol, line 1: Opened an issue up about cleaning up some of these build flags! May be good to get someone familiar with the build system/infrastructure.

kylekrol:
Reworking to split up  the commits in a cleaner fashion

PR #245, opened 2020-09-28
Tie names directly to parameters and state fields to facilitate better error messages

### LINKED ISSUE #232: Improve State Field and Parameter Error Messages
@tanishqaggarwal raised a concern that some of the errors messages produced by failed state field/parameter casts are not very informative:

    auto *field_ptr = dynamic_cast<StateField<T> *>(this);
    if (!field_ptr)
      throw std::runtime_error("Invalid field type in call to 'cast'.");

It would be far superior if this message reported the name of the field the cast failed on. To implement this, we need to tie a string name to the field/parameter itself as opposed to the state entry. This will involve a non-trivial amount of refactoring but shouldn't affect existing model implementations.

kylekrol:
Closes #232.

# Overview

The main goal here was to provide better error messages on failed casts and naming collisions that may be seen in the simulation state. To do so, both parameters and state fields now inherit from `Nameable` which carries a string representation of that particular object's name and a string representation of the type. These strings are now used to generate superior error messages on failed casts and naming collisions.

As an aside, I removed type casting directly on parameters and state fields in favor of requiring users to explicitly call `get` for reads _and_ writes to the underlying value.

The autocoder and Python bindings were updated as needed and the single orbit simulation still runs:

psim


src/psim/core/state.cpp
```diff
@@ -40,18 +40,64 @@ bool State::has_writable(std::string const &name) const {
...
+  {
+    // Check if we have a name collision with existing readable fields
+    auto const iter = _readable_fields.find(field->name());
+    if (iter != _readable_fields.end()) {
+      auto const *_field = iter->second;
+      throw std::runtime_error(
+          "'" + _field->name() + ":" + _field->type() + "' already registered " +
+          "as a readable field. Cannot add '" + field->name() + ":" +
+          field->type() + "' as a writable field"
+        );
```
kkrol27, line 53: Here's an example of what some of the new error messages look like just to give an idea.

include/psim/core/castable.hpp
```diff
@@ -0,0 +1,93 @@
...
+   *
+   *  @{
+   */
+  template <typename U = T>
+  D<U> const &cast() const {
+    static_assert(std::is_same<U, T>::value, "Invalid type in call to Castable::cast() const");
+
+    auto const *ptr = dynamic_cast<D<U> const *>(this);
+    if (!ptr)
+      throw std::runtime_error("Invalid call to 'Castable::cast() const' on '" + type() + ":" + name() + "'");
```
kkrol27, line 73: And error messages for failed dynamic casts.

include/psim/core/castable.hpp
```diff
@@ -0,0 +1,93 @@
...
+   *
+   *  @{
+   */
+  template <typename U = T>
+  D<U> const &cast() const {
+    static_assert(std::is_same<U, T>::value, "Invalid type in call to Castable::cast() const");
+
+    auto const *ptr = dynamic_cast<D<U> const *>(this);
+    if (!ptr)
+      throw std::runtime_error("Invalid call to 'Castable::cast() const' on '" + name() + ":" + type() + "'");
```
kylekrol, line 73: Here is an example of an improved error message on an improper dynamic cast (whether on a parameter or state field).

src/psim/core/state.cpp
kylekrol, line 53: And an improved error message on a naming collision in the simulation state.

tanishqaggarwal (approved):
Looks good

PR #246, opened 2020-09-29
Simulation, simulation runner, and the plugin system for PSim standalone

### LINKED ISSUE #237: Plotting Utility for Standalone PSim Simulations

kylekrol:
Closes #237.

This is the last PSim infrastructure PR! (hopefully at least)

# Overview

Opening this up as a draft for the time being so we can have more eyes on this before it's the architecture we go with. What I'm going for here is to develop two main entry points an end user of PSim can interact with.

The first is just the raw `Simulation` class. I see this as the entrypoint for PTest so PTest can "own" the sim as opposed to the other way around.

The second is the simulation runner and plugin system. In this case, the `SimulationRunner` class contains the "main loop" and the end user just kicks it off. We'll then write plugins to do extra interaction with the simulation such as different stopping conditions, data logging, plotting, and even plugging in control algorithms to perform some tests. Currently, there is a simple stop condition plugin that just looks at step count and a basic 2D/3D data driven plotting tool.

Both of these could be improved in the future.


python/psim/plugins.py
```diff
@@ -0,0 +1,55 @@
...
+        self._steps = 0
+
+    def poststep(self, sim):
+        super(RunForNSteps, self).poststep(sim)
+
+        if self._steps >= self.n
+            sim.should_stop()
+            return
+
+        self._steps = self._steps + 1
```
kylekrol: Here a _very_ simple example of what a plugin could do. Something that would be a little more complicated is a plugin acting as the attitude estimator and one acting as the attitude controller.

tanishqaggarwal (approved):
Looks good.

python/psim/plugins/plot.py
```diff
@@ -0,0 +1,205 @@
...
+        # Ignore plots if the step was set to zero
+        if not self._step > 0:
+            log.warning('Skipping plotting because the plot step wasn\'t greater than zero')
+            return
+
+        log.info('Generating plots...')
+
+        # Generate direct data arrays for plotting
+        arrays = dict()
+        for _arrays in [plot.arrays for plot in self._plots]:
```
tanishqaggarwal, line 182: Man I wrote much of the same code for the plotter on the ptest side...we could've had the room for a lot of reuse here...

tanishqaggarwal (approved):

python/psim/plugins/plot.py
kylekrol: Yeah I was looking at that code when i was writing this. Would've been cool to integrate the telemetry/plotting systems together somehow.

Doesn't mean we can't in the future but I was just looking for the path of least resistance to get something usable up and running.

PR #249, opened 2020-10-01
Add template Js

shihaocao:
# Add template Js

Fixes # N/A
### Summary of changes
- Add template docked sats J

### Ptest Effects
Should be none

### Testing
Just a new constant

### Constants

uhhhhhh do we know the real value right now?

### Documentation Evidence
Should be self documenting


kylekrol (changes requested):
One thing is that all the constants are actually determined in `tools/constants.csv` and PIO will autocode the constants files at build time. Also, I think we decided against have a docked sats moment of inertia constant hardcoded in the satellites?

shihaocao:
Closing since we dont need it

PR #254, opened 2020-10-11
Dual Orbit Simulation

kylekrol:
Small PR to add a simulation where we model two satellites and their orbital mechanics. No attitude dynamics or attitude related actuators are included.

_This small set of changes steps through the infrastructure steps to create a new simulation and make it available in Python so it'd be very beneficial to look through what's going on here!_

### DIFF
--- a/python/psim/_psim/simulations.cpp
+++ b/python/psim/_psim/simulations.cpp
@@ -31,6 +31,7 @@
 #include <psim/core/simulation.hpp>
 #include <psim/core/state_field.hpp>
 #include <psim/core/types.hpp>
+#include <psim/simulations/dual_orbit.hpp>
 #include <psim/simulations/single_orbit.hpp>
 
 #include <lin/core.hpp>
@@ -101,4 +102,5 @@ static void py_assign(psim::StateFieldWritableBase &field, T const &value) {
 
 void py_simulation(py::module &m) {
   PY_SIMULATION(SingleOrbitGnc);
+  PY_SIMULATION(DualOrbitGnc);
 }
--- a/python/psim/_psim/simulations.cpp
+++ b/python/psim/_psim/simulations.cpp
@@ -77,25 +77,25 @@ template <class C>
 class PySimulation : public psim::Simulation {
  public:
   PySimulation() = delete;
-  virtual PySimulation() = default;
+  virtual ~PySimulation() = default;
 
   template <typename... Ts>
-  PySimulation(Configuration const &config, Ts &&... ts)
+  PySimulation(psim::Configuration const &config, Ts &&... ts)
     : Simulation(std::make_unique<C>(config, std::forward<Ts>(ts)...)) { }
 };
 
 #define PY_SIMULATION(model) \
-    py::class_<PySimulation<model>>(m, #model) \
+    py::class_<PySimulation<psim::model>>(m, #model) \
       .def(py::init([](psim::Configuration const &config) { \
-        return new PySimulation<model>(config); \
+        return new PySimulation<psim::model>(config); \
       })) \
-      .def("__getitem__", [](psim::Simulation const &self, std::string const &name) { \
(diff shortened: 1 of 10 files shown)


src/psim/simulations/single_orbit.cpp
```diff
@@ -41,11 +41,13 @@ SingleOrbitGnc::SingleOrbitGnc(Configuration const &config) {
...
-  // Orbital dynamics and lazy transformations
+  // Orbital dynamics
   add<OrbitGncEci>(config, "truth", "leader");
   add<TransformPositionEci>(config, "truth", "truth.leader.orbit.r");
   add<TransformVelocityEci>(config, "truth", "leader", "truth.leader.orbit.v");
-  // Environmental variables with body frame independent lazy transformations
+  // Environmental models
   add<EnvironmentGnc>(config, "truth", "leader");
+  add<TransformPositionEcef>(config, "truth", "truth.leader.environment.b");
+  add<TransformPositionEci>(config, "truth", "truth.leader.environment.s");
```
kylekrol: These two lines here make the following transformations:

 * `truth.leader.environment.b` into `truth.eader.environment.b.eci` and `truth.eader.environment.b.ecef`
 * `truth.eader.environment.s` into `truth.eader.environment.s.eci` and `truth.eader.environment.s.ecef`

python/psim/sims.py
```diff
@@ -2,5 +2,6 @@
 """
 
 from _psim import (
+    DualOrbitGnc,
```
kylekrol, line 5: Here is where we make the new simulation type available for use in Python.

python/psim/_psim/simulations.cpp
```diff
@@ -101,4 +102,5 @@ static void py_assign(psim::StateFieldWritableBase &field, T const &value) {
 
 void py_simulation(py::module &m) {
   PY_SIMULATION(SingleOrbitGnc);
+  PY_SIMULATION(DualOrbitGnc);
```
kylekrol, line 116: Wrapping the new model as a simulation available in Python.

src/psim/simulations/dual_orbit.cpp
```diff
@@ -0,0 +1,60 @@
...
+#include <psim/truth/orbit.hpp>
+#include <psim/truth/time.hpp>
+#include <psim/truth/transform_position.hpp>
+#include <psim/truth/transform_velocity.hpp>
+
+#include <string>
+
+namespace psim {
+
+DualOrbitGnc::DualOrbitGnc(Configuration const &config) {
```
tanishqaggarwal, line 39: Seems like it should be possible to make this code reusable for the SingleOrbit case. As sensor models, etc. develop I'm worried about a lot of code copying occurring across the single and dual-orbit GNC.

tanishqaggarwal (changes requested):

src/psim/simulations/dual_orbit.cpp
kylekrol: This is a good point. I'll break it out into a `SatelliteOrbitGnc` model that's shared across both cases.

python/psim/_psim/simulations.cpp
```diff
@@ -72,18 +73,29 @@ static void py_assign(psim::StateFieldWritableBase &field, T const &value) {
...
+template <class C>
+class PySimulation : public psim::Simulation {
+ public:
+  PySimulation() = delete;
+  virtual ~PySimulation() = default;
+
+  template <typename... Ts>
+  PySimulation(psim::Configuration const &config, Ts &&... ts)
+    : Simulation(std::make_unique<C>(config, std::forward<Ts>(ts)...)) { }
+};
```
kylekrol, line 85: This allows different simulations to be exposed as different objects in Python.

tanishqaggarwal (approved):
Looks like a solid restructure.

PR #255, opened 2020-10-11
Add Sign Usage

shihaocao:
# Add Sign Usage

Fixes #N/A 

### Summary of changes
- Add lin::sign usage

### Ptest Effects
N/A

### Testing
will be tested in fsw

### Constants
None

### Documentation Evidence
Self documenting


kylekrol (commented):
Just be to be thorough, do we want to update line 39 to be:

    if (!lin::all(lin::isfinite(data.b_body))) return;

You'll need to pull in `#include <lin/queries.hpp>` and `#include <lin/math.hpp>` at the top.

tanishqaggarwal (approved):

kylekrol (approved):

PR #260, opened 2020-10-21
Refactor PSim as a Python Module

### LINKED ISSUE #258: Package PSim for Integration with PTest
This will involve two main tasks:

* Packaging PSim in a form that's compatible with `pip install -e`
* Add a `PSIM_CONFIG_PATH` environment variables to allow for easy use of PSim configs from PTest.

kylekrol:
Closes #258.

# Refactor PSim as a Python Module

The two main changes to support PSim integration are:

* Adding a `setup.py` script to build and package the `psim` Python module.
* Adding a `PSIM_CONFIG_PATH` environment variables to allow us to point the simulation to the root of the PSim repository if we're calling it from outside the repository (i.e. in a PTest case)

## Usage

With a virtual environment setup in the PSim repository, you can install the `psim` package from anywhere with the following:

    pip install -e <path-to-psim>

The `-e` option is recommended so Bazel builds in place. This will reduce you're build times and prevent pip from going off and making what Bazel sees as a brand new repository.

With everything installed, PSim standalone can be called with:

    python -m psim --help

and you can set a custom config path as shown below:

    env PSIM_PATH=<path to psim repo> python -m psim --help


tanishqaggarwal (approved):

setup.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+class BuildExtCommand(setuptools.command.build_ext.build_ext):
+
+    def run(self):
+        p = subprocess.Popen(['bazel', 'build', '//:_psim'])
+        while True:
+            try:
+                p.wait()
+                break
```
kylekrol: This used to be `return p.wait()` so of course it wasn't going to work lmao

Issue #224, opened 2020-05-11
Package the Orbit Controller in MATLAB into a Clean Interface


kylekrol:
More details to come!

PR #262, opened 2020-10-25
Orbit Controller and MATLAB Testing

### LINKED ISSUE #222: Orbit Controller Implementation in C++

kylekrol:
# Orbit Controller and MATLAB Testing

Fixes #222, #223, #224.

### Summary of changes
- Added a C++ implementation of the orbit controller.
- Added MEX wrapper.
- MATLAB testing script.

### Ptest Effects
NA

### Testing
Plots from our testing:

pos
vel
h
dv

### Constants
NA

### Documentation Evidence
NA


MATLAB/orbit_controller/conversion.hpp
```diff
@@ -0,0 +1,53 @@
+#include "mex.hpp"
```
fatimayousuf, line 1: This is exactly the same file shihao used for the attitude controller. I don't suppose its a good idea to have two copies of the same file, especially if you all plan on doing similar stuff for other gnc controllers. I can move it so that the attitude+orbit controller mex wrappers both include the same file. it might not be that big of a deal, idk what do you think?

src/gnc/orbit_controller.cpp
```diff
@@ -25,56 +25,188 @@
...
+    double const theta = lin::atan2(
+        lin::dot(that_h_proj, lin::cross(this_r_ecef0, this_h_ecef0)),
+        lin::dot(that_h_proj, this_h_ecef0)
+      ); // ^^ this seems a tad weird to me and would be of very different
+         // orders of magnitude - i.e. `this_r_ecef0` or `this_r_hat`?
+
+    dv_plane = this_h_hat * (K_h * theta / mass);
+    // Don't understand why we divide by mass ^^ as we never were working with
+    // orbital angular momentum anyway
+  }
```
fatimayousuf, line 187: yeah I just copied whatever Stewart did in his test file lol. I think @kkrol27 you and I were talking about this, but I could not really follow what he was doing either haha. but based on the graphs generated, whatever he's doing seems to work, so I didn't question it too much. but if any of you have a better way to do this, I'm all ears

kylekrol: I'm not too worried about it either as it may just be a technicality that changes our gain for this term. Just left it in there in case people had the same question in the future.

tanishqaggarwal (approved):
Overall structure of code looks okay to me. I don't have a great understanding of the physics of the controller but hopefully we can write a paper or two later

PR #263, opened 2020-10-29
Attitude Orbit Simulations for PSim Integration

shihaocao:
# Attitude Orbit Simulations for PTest Integration

The main features added by this PR are:

* `DualAttitudeOrbitGnc` and `SingleAttitudeOrbitGnc` simulation types/models.
* Added a `HillFrame` model interface to support relative position and velocity calculations.
* Basic initial conditions for a deployment case (see `config/parameters/truth/deployment.txt`).
* Small update(s) to the attitude orbit model:
  * Lazy evaluation quaternion function.
  * Use body frame magnetic field instead of intertial frame.
* Other random bug fixes throughout PSim involving infinite recursion causing a segfault, a coordinate frame issue, and other miscelaneous issues.
* Initial conditions for deployment and detumble testcases.

Note for future debugging, it's possible to just run `gdb python` and the call the following from inside `gdb` to get a stack trace out of PSim on assertion/segfault issues:

    gdb python
    (gdb) r -m psim <arguments producing an error>

## Verification

We were able to successfully detumble in PSim standalone:

detumble

The detumbler for PSim standalone isn't included here and will be included with a seprate PR as it was super hacky.

@shihaocao can also confirm this version of PSim allowed us to run a successful detumble case in HOOTL!


config/parameters/detumble.txt
```diff
@@ -0,0 +1,26 @@
...
+truth.leader.orbit.v  -7236.19   -1695.2    -1708.76
+
+# Follower spacecraft attitude and orbit initial conditions.
+
+truth.follower.orbit.r  -2.16425e6  4.60331e6  4.59251e6
+truth.follower.orbit.v  -7236.17   -1695.04   -1708.5
+
+truth.follower.attitude.q.body_eci -0.282557 -0.604902  -0.71832   -0.195632
+truth.follower.attitude.w          -0.247035  0.0574472  0.0659878
+truth.follower.wheels.w             0.0       0.0        0.0
```
kylekrol: Initial conditions for a detumble case. These were generated form a deployment case using the snapshot feature with the Python command listed in the file's header.

config/parameters/truth/base.txt
```diff
@@ -0,0 +1,23 @@
...
+truth.leader.wheels.J      135.0e-7
+truth.leader.wheels.w_max  677.0
+
+# Follower spacecraft attitude and orbit shared initial conditions.
+
+truth.follower.m  5.0
+truth.follower.J  0.03798 0.03957 0.00688
+
+truth.follower.wheels.J      135.0e-7
+truth.follower.wheels.w_max  677.0
```
kylekrol, line 23: Set of configuration constants shared across all simulations.

config/parameters/truth/deployment.txt
```diff
@@ -0,0 +1,27 @@
...
+#
+# The velocity vector is slightly perturbed from the leader's to allow them to
+# drift apart.
+
+truth.follower.attitude.q.body_eci  0.0  0.0  0.0 1.0
+truth.follower.attitude.w           0.2 -0.15 0.1
+truth.follower.wheels.w             0.0  0.0  0.0
+
+truth.follower.orbit.r  6.8538e6 0.0      0.0
+truth.follower.orbit.v  0.0      5.3954e3 5.3951e3
```
kylekrol, line 27: Deployment initial conditions where the satellites start at the same location with a small relative velocity and different angular rates.

include/psim/truth/attitude_orbit.yml
```diff
@@ -84,5 +88,5 @@ gets:
       type: Real
     - name: "{prefix}.earth.q.eci_ecef"
       type: Vector4
-    - name: "{prefix}.{satellite}.environment.b.eci"
+    - name: "{prefix}.{satellite}.environment.b.body"
```
kylekrol, line 91: We needed magnetic field in the body frame not ECI - my bad originally.

include/psim/truth/attitude_orbit.yml
```diff
@@ -76,6 +76,10 @@ adds:
       comment: >
           Magnetic moment commanded of the magnetic torque rods. This value is
           defaulted back to zero at each step.
+    - name: "{prefix}.{satellite}.attitude.L"
+      type: Lazy Vector3
+      comment: >
+          Angular momentum of the spacecraft in the body frame.
```
kylekrol, line 82: Added a lazily evaluated angular momentum field for debugging.

python/psim/simulation.py
```diff
@@ -53,6 +53,15 @@ def __setitem__(self, name, value):
...
 
+    def get(self, name):
+        """Attempts to retrieve a state field from the underlying simulation.
+        Upon failure, 'None' is returned.
+        """
+        try:
+            return self[name]
+        except RuntimeError:
+            return None
+
```
kylekrol, line 64: Added a way to get a field without potentially raising an exception.

src/psim/truth/earth.cpp
```diff
@@ -42,7 +42,7 @@ Vector4 EarthGnc::prefix_earth_q_ecef_eci() const {
 }
 
 Vector4 EarthGnc::prefix_earth_q_eci_ecef() const {
-  auto const &q_ecef_eci = this->Super::prefix_earth_q_eci_ecef.get();
+  auto const &q_ecef_eci = this->Super::prefix_earth_q_ecef_eci.get();
```
kylekrol, line 45: This was the bug causing infinite recursion and a segfault!

src/psim/truth/transform_direction.cpp
```diff
@@ -47,19 +47,19 @@ Vector3 TransformDirectionBody::vector_ecef() const {
...
 }
 
 Vector3 TransformDirectionEcef::vector_body() const {
   auto const &d_eci = Super::vector_eci.get();
-  auto const q_eci_body = prefix_satellite_attitude_q_eci_body->get();
+  auto const q_body_eci = prefix_satellite_attitude_q_body_eci->get();
 
   Vector3 d_body;
-  gnc::utl::rotate_frame(q_eci_body, d_eci, d_body);
+  gnc::utl::rotate_frame(q_body_eci, d_eci, d_body);
```
kylekrol, line 62: These were breaking the attitude estimator as they're incorrectly converting the magnetic field reading to the body frame.

config/parameters/truth/base.txt
tanishqaggarwal: A lot of these constants look like they'd also be part of the existing C++ GNC code. Can we use those for configuration?

config/parameters/truth/detumble.txt
```diff
@@ -0,0 +1,26 @@
+# Initial conditions for simulations starting during detumble. These initial
+# conditions were generated with the following command:
+#   python -m psim -s 10000 -c truth/base,truth/deployment --snapshot snapshot.txt DualAttitudeOrbitGnc
+#
+
+# Time model initial conditions.
+
+truth.t.ns  1700000000000
```
tanishqaggarwal, line 8: This parameter will change if the flight software cycle time changes (which is possible.) Tracked in https://github.com/pathfinder-for-autonomous-navigation/FlightSoftware/issues/532

python/psim/plugins/snapshot.py
```diff
@@ -0,0 +1,65 @@
...
+        self._snapshot = args.snapshot
+        log.info('Saving simulation snapshot to "%s" upon simulation termination.', self._snapshot)
+
+    def cleanup(self, sim):
+        super(Snapshot, self).cleanup(sim)
+
+        if not self._snapshot:
+            return
+
+        _fields = [
```
tanishqaggarwal, line 48: Would there be any reason for a list like this to be passed in as, say, a filename via a psim command line option? Like do we expect this set of parameters to change? If so, might be worth putting in the work now to have it work?

python/psim/simulation.py
tanishqaggarwal: Is this for the PTest integration? Isn't it dangerous for stuff like this to fail silently?

src/psim/simulations/single_orbit.cpp
```diff
@@ -22,7 +22,7 @@
 // SOFTWARE.
```
tanishqaggarwal, line 22: What is the reason for keeping this model around, now that the attitude-orbit model exists?

src/psim/truth/transform_direction.cpp
tanishqaggarwal: What is our naming convention? Does `q_body_eci` mean the rotation from ECI to body, or the other way around?

tanishqaggarwal (approved):
Hooray!!

python/psim/simulation.py
kylekrol: This is used, right now, for the snapshot feature that dumps the current state of the simulation upon termination and not in PTest as far as I'm aware.

src/psim/truth/transform_direction.cpp
kylekrol: `q_a_b` means a  rotation from `b` to `a`.

python/psim/plugins/snapshot.py
kylekrol: Given that we'd only be tracking truth data here, no I don't foresee this changing too much. Also, the feature is only used to generate new test cases so from the perspective I don't think the extra work is really worth it atm.

src/psim/simulations/single_orbit.cpp
kylekrol: If I end up running orbit estimator testing and don't want/need all the attitude simulation stuff involved. This is actually exactly why I wanted to shift to this type of architecture!

config/parameters/truth/detumble.txt
kylekrol: Thanks - and yeah @shihaocao will be keeping track of that for now. #266 will help make changes possible from the PTest side as well.

config/parameters/truth/base.txt
kylekrol: The constants are in the existing GNC code. However, in order to have the "truth" constants separate from flight software's set of constants we should keep them in this file separately. It'll allow us to vet the constants going to make sure the system doesn't exhibit vastly different behavior given small perturbations in some of the constants.

It's unfortunate that they're are two sets but there should be a separate set for the truth and flight software implementations.

PR #269, opened 2020-11-05
Added Skeleton Sensor Models and Removed the Prefix Argument

kylekrol:
## Added Skeleton Sensor Models and Removed the Prefix Argument

* Added "perfect" sensor models for the GPS, gyroscope, magnetometer, and sun sensors. There is still a lot of improvement to be made here as the sensor models don't really _do_ anything. However, it'll be nice to have the fields in PSim earlier rather than later.
* Removed the `prefix` argument from most truth models. All truth models were prefixed with "truth" so I didn't see the utility in keeping the extra model argument floating around and convoluting things.

This _should_ truly be the last of the massive PSim PRs - very exciting!


BUILD.bazel
```diff
@@ -52,6 +52,20 @@ psim_cc_library(
     visibility = ["//visibility:public"],
 )
 
+# Builds the PSim flight computer models
+# psim_autocoded_cc_library(
```
tanishqaggarwal: Why is a flight controller model necessary, again? I thought the whole point of the Python integration was to avoid having to sim the flight controller.

tanishqaggarwal: Why is this commented out?

include/psim/sensors/gps_no_attitude.yml
```diff
@@ -0,0 +1,23 @@
+
+name: GpsNoAttitudeInterface
```
tanishqaggarwal, line 2: Why the "NoAttitude" in the name?

include/psim/sensors/magnetometer.yml
```diff
@@ -0,0 +1,24 @@
+
```
tanishqaggarwal, line 1: Should there be another prefix for magnetometer #? since we have 2?

include/psim/sensors/sun_sensors.hpp
```diff
@@ -0,0 +1,50 @@
+//
```
tanishqaggarwal, line 1: Should we go a level deeper and make a sensor model for each sensor?

tanishqaggarwal (approved):

include/psim/sensors/sun_sensors.hpp
kylekrol: This is a completely valid concern and something that could be improved upon in the future. For the time being, I just wanted to `sensors` model to have this field in here. If how the sun vector is generated is completely changed later is a whole other story.

include/psim/sensors/magnetometer.yml
kylekrol: I would say no to this as both magnetometers should behave similarly and their difference is abstracted away at a fairly lower level. That being said, like the sun sensors, this is something that could approved upon in the future.

include/psim/sensors/gps_no_attitude.yml
kylekrol: We will want a GPS model that doesn't rely on attitude dynamics for simulations without attitude dynamics. We also can in the future add in a GPS model that looses signal if we aren't pointing in the right direction (i.e. a model just called `GpsInterface`).

BUILD.bazel
kylekrol: Ah sorry I will take this out of this PR! And how we want to handle PSim standalone is still a fluid topic but sticking to C++ is the lowest barrier of entry in terms of using our GNC algorithms directly.

BUILD.bazel
```diff
@@ -52,6 +52,20 @@ psim_cc_library(
...
+#     name = "fc",
+#     deps = ["@lin//:lin", "//:gnc", "//:psim_core", "//:psim_utilities"],
+#     visibility = ["//visibility:public"],
+# )
```
kkrol27, line 60:
```suggestion
```

BUILD.bazel
```diff
@@ -52,6 +52,20 @@ psim_cc_library(
...
+
```
kylekrol:
```suggestion
```

PR #272, opened 2020-11-12
Remove Virtual Environment Requirement from Bazel Builds

kylekrol:
Previously, a local virtual environment had to exit with `pyyaml` installed for the autocoder to run. I've removed that dependency in favor on the experimental Python rules handling of pip requirements.

This will make installing with `pip install -e <path-to-psim>` easier as a virtual environment won't have to be present for the install to work!


kylekrol:
@shihaocao this means on the flight software side of things simply cloning the psim submodule and calling `pip install -e lib/common/psim` will now work!

tanishqaggarwal (approved):

shihaocao:
bet

PR #275, opened 2020-11-19
Make Simulation Configuration Editable from Python

### LINKED ISSUE #266: Make Configuration Values Editable in Python
Essentially, especially in `ptest`, we would like to be able to edit some initial conditions prior to running cases. This requires making small changes to the `psim::Configuration` class and it's corresponding Python bindings.

kylekrol:
Fixes #266.

# Make Simulation Configuration Editable from Python

Per the title, this change makes a simulation configuration object editable from Python. You can add and/or overwrite items in the configuration to whatever you prefer. Do note, however, that you can change the type of the underlying parameter as well!

I've also slightly refactored the `Simulation` interface - heads up on that @shihaocao. Now, you create the configuratoin yourself and pass it as an argument to the simulation like this:

    import lin, psim
    config = psim.Configuration(['config/deployment.txt', ...])
    sim = psim.Simulation(psim.sims.Type, config)

Here is a small test I did in the Python shell:

    @thinkpad ~/git/pan/psim git:(feature/update-parameters|+1) venv:(3.8.6)
    % python
    Python 3.8.6 (default, Sep 30 2020, 04:00:38)
    [GCC 10.2.0] on linux
    Type "help", "copyright", "credits" or "license" for more information.
    >>> import lin, psim
    >>> config = psim.Configuration()
    >>> config['test.integer'] = 5
    >>> config['test.integer']
    5
    >>> exit()

I also ran a single satellite attitude orbit case locally to verify the change!

### DIFF
--- a/include/psim/core/configuration.hpp
+++ b/include/psim/core/configuration.hpp
@@ -35,44 +35,49 @@
 #include "types.hpp"
 
 #include <functional>
+#include <memory>
 #include <string>
 #include <unordered_map>
 #include <vector>
 
 namespace psim {
 
-/** @brief Represents a set of parameters used to initialize a simulation.
- *
- *  A configuration is a set of parameters parsed from a configuration file.
- *  These parameters are made accesible to the simulation's models during
- *  intialization to specify initial conditions and configure constants.
- */
 class Configuration {
  private:
-  /** @brief Map containing the parameter pointers.
-   *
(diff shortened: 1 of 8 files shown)


kylekrol:
@shihaocao Once this is in we can move to clean up some of the parameter setting on the PTest side of things! Perhaps add a section to the psim configuration files specifying overloads; althought, handling it on a case by case basis may be better because the changes will be smaller.

shihaocao (approved):

shihaocao:
fat bet

PR #277, opened 2020-11-26
Seeded Random Number Generation in PSim

### LINKED ISSUE #252: Add a random seed argument to PSim
We need to add a way to pass a random seed to PSim. This isn't absolutely critical at the moment as random seeding in `lin` is deterministic with default arguments but we'll want to add flexibility in the future when we're handling Monte Carlo simulations and what not.

This may require adding a `lin::internal::RandomsGenerator` to the simulation class itself.

kylekrol:
Closes #252.

# Seeded Random Number Generation in PSim

Per the title, seeded random number generation is now available in PSim. The random number generator is owned by the simulation class and is therefore seeded in each individual instantiation of a simulation - very helpful if we want to run multiple sims in parallel at some point. Each model class then has access to the random number generator via the `_randoms` member variable shared across all models.

You can generate random numbers using the standard `lin` utilities:

```
#include <lin/generators/randoms.hpp>

... {
  // To make lin objects:
  auto gaussians = lin::gaussians<...>(_randoms,...);
  auto randoms = lin::rands<...>(_randoms,...);
  // Just a single scalar
  auto gaussian = _randoms.gaussian();
  auto random = _randoms.rand();
}
```

I reran simulation locally to verify the change wasn't breaking. You can also see the random number generator in action in #278.


include/psim/core/simulation.hpp
```diff
@@ -32,54 +32,53 @@
...
+  C _model;
 
  public:
   Simulation() = delete;
   Simulation(Simulation const &) = delete;
-  Simulation(Simulation &&) = default;
+  Simulation(Simulation &&) = delete;
   Simulation &operator=(Simulation const &) = delete;
-  Simulation &operator=(Simulation &&) = default;
+  Simulation &operator=(Simulation &&) = delete;
```
kylekrol, line 61: These are deleted now so that the references to the randoms generator in the model isn't invalidated by a move or copy.

include/psim/core/simulation.hpp
```diff
@@ -32,54 +32,53 @@
...
    */
-  std::unique_ptr<Model> _model;
+  RandomsGenerator _randoms;
 
- protected:
-  /** @brief Private constructor employed by the factory function.
+  /** @brief Model employed by the simulation.
    */
-  Simulation(std::unique_ptr<Model> &&model);
+  C _model;
```
kylekrol, line 54: No need to be stored in a unique pointer if the class can't be moved.

tools/autocoder.py
```diff
@@ -372,14 +372,14 @@ def code(self):
...
-            '  {}(Configuration const &config'.format(self._name)
+            '  {}(RandomsGenerator &randoms, Configuration const &config'.format(self._name)
 
             # Arg arguments
             for arg in self._args:
                 self.__code += ', std::string const &' + arg._name
 
             self.__code += ')\n' + \
-            '  : {}()'.format(self._type)
+            '  : {}(randoms)'.format(self._type)
```
kylekrol, line 382: These are the two changes required to the auto-coded header files.

kylekrol:
Currently block on this quick PR: https://github.com/pathfinder-for-autonomous-navigation/lin/pull/17

tanishqaggarwal (approved):
Changes seem innocuous enough

shihaocao (approved):

PR #278, opened 2020-11-27
GPS, Gyroscope, and Magnetometer Sensor Models

kylekrol:
# GPS, Gyroscope, and Magnetometer Sensor Models

Simple sensors models were implemented for the three sensors specified in the title above. Both the GPS and magnetometer models are just white noise models at the moment. We should moving forward add in dependence on the MTRs for the magnetometer model.

The gyroscope model is based off of Stewarts model he used during development of the attitude filter. It's a white noise model added on top of a hard offset (bias) that randomly walks over time.

All of the noise parameters are configurable via the simulation config prior to initialization!

I've attached some plots below showing the sensor model's error over time for the gyroscope and magnetometer. The command ran to obtain them was:

    python -m psim -s 1000 -p sensors/gps,sensors/gyroscope,sensors/magnetometer -ps 1 -c sensors/base,truth/base,truth/deployment SingleAttitudeOrbitGnc

gyr
gyr_err
mag
mag_err

### DIFF
--- a/config/parameters/sensors/base.txt
+++ b/config/parameters/sensors/base.txt
@@ -1,24 +1,20 @@
 # Basic configuration information shared across all sensor model initial
 # conditions.
 #
+# All configuration values for the gyroscope and magnetometer were
+# experimentally determined. See the following folder in the PAN team google
+# drive:
+#   System Level > Integration and Testing > Sensor Charactarization
+#
 
 # Leader spacecraft base sensor configuration
 
 sensors.leader.gps.r.sigma  0.0 0.0 0.0
 
-# The source for these numbers is currectly Stewart's original attitude
-# estimator script written in MATLAB. This will be updated down the road
-# but the datasheet is still linked below.
-#
-# https://www.st.com/content/ccc/resource/technical/document/datasheet/76/27/cf/88/c5/03/42/6b/DM00218116.pdf/files/DM00218116.pdf/jcr:content/translations/en.DM00218116.pdf
 sensors.leader.gyroscope.w.bias        0.02    0.01    -0.03
 sensors.leader.gyroscope.w.bias.sigma  1.00e-6 1.00e-6  1.00e-6
 sensors.leader.gyroscope.w.sigma       2.75e-4 2.75e-4  2.75e-4
 
-# Pulled from the LIS3MDL datasheet. Note that the noises specified there are
-# 3.2e-, 3.2e-7, and 4.0e-7 respectively but we are being conservative here.
-#
-# https://www.st.com/resource/en/datasheet/lis3mdl.pdf
 sensors.leader.magnetometer.b.sigma  5.00e-7 5.00e-7 5.00e-7
 
 # Follower spacecraft base sensor configuration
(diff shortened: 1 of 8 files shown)


kylekrol:
This PR will be rebased onto master once #277 is merged.

config/parameters/sensors/base.txt
```diff
@@ -1,10 +1,23 @@
...
 # Leader spacecraft base sensor configuration
 
-sensors.leader.gyroscope.bias  0.02 0.01 -0.03
+sensors.leader.gps.r.sigma  0.0 0.0 0.0
+
+sensors.leader.gyroscope.w.bias        0.02    0.01    -0.03
+sensors.leader.gyroscope.w.bias.sigma  1.00e-6 1.00e-6  1.00e-6
+sensors.leader.gyroscope.w.sigma       2.75e-4 2.75e-4  2.75e-4
+
+sensors.leader.magnetometer.b.sigma  5.00e-7 5.00e-7 5.00e-7
```
tanishqaggarwal, line 18: What are the sources for these numbers? Should we link them in this config file so that the source is known?

src/psim/sensors/gyroscope.cpp
```diff
@@ -33,15 +33,19 @@ namespace psim {
...
 
 Vector3 Gyroscope::sensors_satellite_gyroscope_w() const {
   auto const &truth_w = truth_satellite_attitude_w->get();
-  auto const &bias = sensors_satellite_gyroscope_bias.get();
+  auto const &bias = sensors_satellite_gyroscope_w_bias.get();
+  auto const &sigma = sensors_satellite_gyroscope_w_sigma.get();
 
-  // TODO : Implement a white noise model
-  return truth_w + bias;
+  return truth_w + bias + lin::multiply(sigma, lin::gaussians<Vector3>(_randoms));
```
tanishqaggarwal, line 48: I imagine this kind of modeling (bias + truth) will be extremely pervasive across many sensor models, should we build some kind of library that will automatically figure out the right state fields to use for offsetting truth?

kylekrol: Believe it or not, this is the only sensor on the spacecraft that has a bias worth modelling. That being said, the generic white noise model is a bit more pervasive in the code base (pretty much every other sensor).

If we try to generalize that model we may be able to. That being said though we'd lose all the sensor specific comments in the YAML file and things wouldn't be documented as well. For that reason I'm hesitant to generalize the sensor model to a single interface with more arguments.

**EDIT:** I would say though this is one downfall of the PSim architecture is there is no semblance of inheritance among models.

config/parameters/sensors/base.txt
kylekrol: Yep, very good point. Currently, my source was Stewart's attitude estimator MATLAB script but I will see if he linked anything there. If not, I may open up a separate ticket about adding links/documentation for these numbers because I have to focus on other school things for the next few days and then will be travelling and may not have time to fix this PR up before then.

shihaocao (approved):
Sweet

src/psim/sensors/gyroscope.cpp
tanishqaggarwal: Let's just leave things as is.

config/parameters/sensors/base.txt
kylekrol: Updated a comment at the top of the file. These values were derived experimentally and the documentation in stored in the listed path on Google drive!

PR #285, opened 2020-12-23
Refactor pybind11 Code to Fix Configuration Bug

kylekrol:
See title for summary. All `pybind11`code was all consolidated to a single file as well to make the Python specific types more visible.

@shihaocao here is some terminal output. I believe the bug has been fixed.

```
@jezero ~/git/pan/psim git:(bugfix/configuration|+7) venv:(3.9.1)
% python
Python 3.9.1 (default, Dec 13 2020, 11:55:53)
[GCC 10.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import lin, psim
>>>
>>> config = psim.Configuration(['config/parameters/truth/base.txt', 'config/parameters/truth/deployment.txt', 'config/parameters/sensors/base.txt'])
>>> config['truth.leader.attitude.w'] = lin.Vector3()
>>> config['truth.leader.attitude.w']
Vector3
0
0
0
>>> exit()
```

### Refactoring of the Configuration Object's Parameter Dictionary

The main issue behind the bug was the fact that the configuration's parameter dictionary used a reference to the parameter's string name as a key. When we overwrote a parameter in the dictionary, the referenced string key from the previous parameter was freed and potentially overwritten making it seem like that key didn't exist in the dictionary anymore. As such, the dictionary in the configuration object was updated to store a string by value as opposed to by reference.

### Migration of `pybind11` Code

The `pybind11` code was also moved to `python/psim/_psim.cpp` as opposed to `src/psim/_psim/**` as it's not a C++ library.


shihaocao:
Rip, unable to confirm

shihaocao:
image

shihaocao:
Changes look good to me. Verified working in PTest :)

shihaocao (approved):

include/psim/core/configuration.hpp
```diff
@@ -56,21 +56,17 @@ class Configuration {
...
  protected:
   /** @brief Map containing the parameter pointers.
-   *
-   *  https://stackoverflow.com/questions/9139748/using-stdreference-wrapper-as-the-key-in-a-stdmap
    */
-  std::unordered_map<std::reference_wrapper<std::string const>,
-                     std::unique_ptr<ParameterBase const>,
-                     std::hash<std::string>, std::equal_to<std::string>>
-                     _parameters;
+  std::unordered_map<std::string, std::unique_ptr<ParameterBase const>>
```
shihaocao, line 60: so this was the critical change related to the bug?

src/psim/core/configuration.cpp
```diff
@@ -44,13 +44,11 @@ void Configuration::_add(std::string const &name, T &&value,
...
-                             "as a configuration file.\n" +
-                             "Duplicate parameter name '" + name + "'.");
+    throw std::runtime_error(
+        "Error on line + " + std::to_string(l) + " while parsing " + file +
+        "as a configuration file.\n" + "Duplicate parameter '" + name + "'.");
 
-  auto param = std::make_unique<Parameter>(name, std::forward<T>(value));
-  _parameters[param->name()] = std::move(param);
+  _parameters[name] = std::make_unique<Parameter>(name, std::forward<T>(value));
 }
```
shihaocao, line 52: along with these lines?

include/psim/core/configuration.hpp
kylekrol: Yes this was absolutely critical. I'm going to add a better explanation in the description of exactly what was going on.

src/psim/core/configuration.cpp
kylekrol: These changes here weren't critical per-say but the lines could be condensed because we didn't have to careful about which string we were using to do the dictionary lookup.

**EDIT:** With the old design, it was imperative the assignment was done with string stored within the parameter object itself as it's taken by reference. This, however, is no longer the case with the update.

PR #287, opened 2021-01-07
Updated PSim Standalone Plotting Utilities

kylekrol:
# Update PSim Standalone Plotting Utilities

### Summary of changes

The `psim` standalone plotting utilities only supported two styles (two and three dimensional plotting). The major issues was the lack of support for log scale plots and estimate plots with two sigma bounds.

The plotting utilities were overhauled accordingly and now feature multiple "plot types" including `Plot2D`, `Plot2DLog`, `Plot3D`, and `PlotEstimate`. There are examples all each being used in the updated plotting files.

### Testing

All plotting utilities were tested locally and some screenshots are included below featuring the updated plots for the attitude estimator.

log_deg_err
px py pz
wx wy wz

All plots were generated with the following command:

    python -m psim -s 1000 -p fc/attitude,truth/attitude,sensors/sun_sensors -ps 1 -c sensors/base,truth/base,truth/deployment -v AttitudeEstimatorTestGnc

PR #280, opened 2020-12-04
Attitude Estimator Model

kylekrol:
# Attitude Estimator Model

Closes #197.

Pairs with #287 and #288.

### Summary of Changes

- Added an `AttitudeEstimator` model to mimic the attitude estimator running on the flight computer. More telemetry is exposed as runtime than is available off the flight computer itself.
- Added a new simulation `AttitudeEstimatorTestGnc` to actually run tests.

Note that most of the lazily evaluated fields simply facilitate checks on filter health. They are designed to mimic the plots @stewartaslan provided in his old MATLAB script.

### Testing

See the plots included with #287 as well as the screenshot below.

everything

PR #288, opened 2021-01-07
Improved Sun Sensors Model

kylekrol:
# Improved Sun Sensors Model

Relates to #143.

### Summary of changes

- Added single spherical angle noise model. This model is currently employed by the filter as well but will have to be updated as the model breaks down when the sun vector is near the z axis.
- Added a simple eclipse model. Note, the model must be enable for it to be active (see config file additions).

### Testing

Some sample images taken with eclipse and without eclipse are included here.

With eclipse enabled:

eclipse_s
eclipse_s_error

Without eclipse enabled:

s
s_error

You can also see the more detailed behavior with a shorter time scale:

short_s
short_s_error

The above plots can be generated with the following command:

    python -m psim -s 20000 -p sensors/sun_sensors -ps 1 -c sensors/base,truth/base,truth/deployment -v SingleAttitudeOrbitGnc

### DIFF
--- a/config/parameters/sensors/base.txt
+++ b/config/parameters/sensors/base.txt
@@ -6,6 +6,12 @@
 # drive:
 #   System Level > Integration and Testing > Sensor Charactarization
 #
+# The noise distribution for the sun sensors was chosen as two degrees, one
+# sigma. This matches the attitude filter's settings.
+#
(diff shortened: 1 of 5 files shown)


kylekrol:
This needs to be rebased once #280 and #287 are merged.

shihaocao (approved):

PR #290, opened 2021-01-15
Update reason for disabling the eclipse model

kylekrol:
Small comment update. Previous comment ended up being incorrect as the filter behaves well during eclipse.


shihaocao (approved):
Bet

PR #293, opened 2021-01-24
Update lin Pointer

kylekrol:
Requires https://github.com/pathfinder-for-autonomous-navigation/lin/pull/22 and https://github.com/pathfinder-for-autonomous-navigation/lin/pull/23 to be merged.

### DIFF
--- a/src/psim/truth/attitude_orbit.cpp
+++ b/src/psim/truth/attitude_orbit.cpp
@@ -78,7 +78,7 @@ void AttitudeOrbitNoFuelEciGnc::step() {
   IntegratorData data {mass, I, I_w, tau_w, m, b};
 
   // Simulate our dynamics
-  lin::Vector<Real, 16> const x = {
+  Vector<16> const x = {
       r(0), r(1), r(2),
       v(0), v(1), v(2),
       q_body_eci(0), q_body_eci(1), q_body_eci(2), q_body_eci(3),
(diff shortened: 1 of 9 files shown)


src/psim/truth/attitude_orbit.cpp
```diff
@@ -87,16 +87,16 @@ void AttitudeOrbitNoFuelEciGnc::step() {
...
-    auto const r = lin::ref<3, 1>(x, 0,  0);
-    auto const v = lin::ref<3, 1>(x, 3,  0);
-    auto const q = lin::ref<4, 1>(x, 6,  0);
-    auto const w = lin::ref<3, 1>(x, 10, 0);
-    auto const w_w = lin::ref<3, 1>(x, 13, 0);
+    auto const r = lin::ref<Vector3>(x, 0,  0);
+    auto const v = lin::ref<Vector3>(x, 3,  0);
+    auto const q = lin::ref<Vector4>(x, 6,  0);
+    auto const w = lin::ref<Vector3>(x, 10, 0);
+    auto const w_w = lin::ref<Vector3>(x, 13, 0);
```
shihaocao, line 96: why not explicitly Vector3d?

shihaocao (approved):
Just a question

src/psim/truth/attitude_orbit.cpp
kylekrol: I added aliases in the `::psim` namespace for `Matrix<...>`, `Vector<...>`, `RowVector<...>`, `Vector2`, `Vector3`, and `Vector4` which all assume double elements. This is because Python by default only uses doubles and not floats so it makes pybind11 code easier to write.

Issue #142, opened 2020-03-11
GPS Sensor Characterization and Models


kylekrol:

kylekrol:
Updating the tags on this issue. There is currently an empty GPS model in psim and we can move forward with implementing a noise model given @shihaocao's addition of `lin::gaussians`.

kylekrol:
GPS sensor noise values were set to 5.0 m and 5.0 m/s one sigma for position and velocity measurements respectively #294. These values have not been researched and should be either confirmed or updated accordingly.

PR #294, opened 2021-01-24
Add GPS Velocity Readout

kylekrol:
Testing required #292.

Relates to #142.

### Summary of Changes

* A velocity readout was added to the GPS to better match the sensor on orbit.
* Uncertainties of `5.0 m` and `5.0 m/s` (both one sigma) were added to the position and velocity readouts of the GPS. These values will be updated properly in #296.

### Testing

See the images below generated via the following command:

    python -m psim -s 2000 -p truth/orbit,sensors/gps -ps 1 -c sensors/base,truth/base,truth/deployment -v SingleOrbitGnc

truth_v
gps_v
gps_v_err

### DIFF
--- a/config/parameters/sensors/base.txt
+++ b/config/parameters/sensors/base.txt
@@ -16,7 +16,8 @@
 
 # Leader spacecraft base sensor configuration
 
-sensors.leader.gps.r.sigma  0.0 0.0 0.0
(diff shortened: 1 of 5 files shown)


shihaocao (approved):
bet

PR #297, opened 2021-02-03
Orbit Library Orbit Estimator Implementation

### LINKED ISSUE #78: Add Orbital State Estimation Function

kylekrol:
Closes #78.

### Summary of changes

- Small tweaks to the `//:gnc` target to avoid loosing headers in nested directories within `include/orb/`.
- Square root extended Kalman filter orbital state estimator implementation. GPS position and velocity readouts are taken as sensor measurements in ECEF.
- Flight computer orbit estimator model for `psim` with plots config included.
- Simple orbit estimator simulation for testing in `psim` standalone.

### Testing

See the following plots generate below showing valid estimates of position and velocity over time. Tests were run with the following command:

    python -m psim -s 10000 -p fc/orbit,sensors/gps -ps 1 -c sensors/base,truth/base,truth/deployment -v OrbOrbitEstimatorTest

rx
ry
rz
vx
vy
vz

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -21,7 +21,7 @@ cc_library(
         "include/gnc/**/*.inl"
     ]),
     hdrs = glob([
-        "include/gnc/**/*.hpp", "include/orb/*.h", "include/orb/*.hpp"
+        "include/gnc/**/*.hpp", "include/orb/**/*.h", "include/orb/**/*.hpp"
     ]),
     includes = ["include"],
     copts = ["-Isrc", "-fvisibility=hidden"],
--- a/config/plots/fc/orbit.yml
+++ b/config/plots/fc/orbit.yml
@@ -1,0 +1,26 @@
+
+# Orbit estimate's validity over time.
+- type: Plot2D
+  x: truth.t.s
+  y: fc.leader.orbit.is_valid
+
+# Orbit estimator's performance in terms of position and velocity estimate
+# error.
+- type: PlotEstimate
+  x: truth.t.s
+  y: fc.leader.orbit.r.x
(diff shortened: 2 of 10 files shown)


kylekrol:
I'd recommend going through the PR commit by commit; the changes will hopefully be a little clearer that way!

BUILD.bazel
```diff
@@ -21,7 +21,7 @@ cc_library(
         "include/gnc/**/*.inl"
     ]),
     hdrs = glob([
-        "include/gnc/**/*.hpp", "include/orb/*.h", "include/orb/*.hpp"
+        "include/gnc/**/*.hpp", "include/orb/**/*.h", "include/orb/**/*.hpp"
```
kylekrol, line 24: This changed allows Bazel to pick up the nested header files in the `orb` library.

afoarce (approved):
Overall looks good to me; being less familiar with the algorithm I mostly looked at structure and output.

Issue #141, opened 2020-03-11
CDGPS Sensor Characterization and Models


kylekrol:

kylekrol:
#302 set the sensor noise values to a somewhat arbitrary value of 10 cm in each directionl; we need to verify this number./

PR #302, opened 2021-02-07
CDGPS Sensor Model

kylekrol:
Relates to #141.

### Summary of changes
- Simple range based CDGPS sensor with additive gaussian noise.
- Sensor plotting files.
- Updated default PSim configurations.
- Added CDGPS sensors to all dual cases.

### Testing

The CDGPS sensor model was tested using deployment initial conditions to verify the range based sensor behavior:

    python -m psim -s 60000 -p sensors/cdgps,sensors/gps -ps 10 -c sensors/base,truth/base,truth/deployment -v DualOrbitGnc

Notice how a little over 8000 seconds in the CDGPS sensor readings disappear where as GPS readings don't; this means the sensor was deactivate at longer ranges as expected.

cdgps
cdgps_error
gps

### DIFF
--- a/include/psim/sensors/cdgps_no_attitude.yml
+++ b/include/psim/sensors/cdgps_no_attitude.yml
@@ -34,12 +34,13 @@ adds:
     - name: "sensors.{satellite}.cdgps.dr"
       type: Lazy Vector3
       comment: >
-          Relative position of the other satellite reported by the CDGPS in ECEF.
+          Relative position this satellite to the other reported by the CDGPS in
+          ECEF.
     - name: "sensors.{satellite}.cdgps.dr.error"
       type: Lazy Vector3
       comment: >
-          Error in the relative position of the other satellite by the CDGPS in
-          ECEF.
+          Error in the relative position of the this satellite to the other
+          satellite by the CDGPS in ECEF.
 
 gets:
     - name: "truth.{satellite}.orbit.r.ecef"
--- a/include/psim/sensors/cdgps_no_attitude.yml
+++ b/include/psim/sensors/cdgps_no_attitude.yml
@@ -1,0 +1,48 @@
+
+name: CdgpsNoAttitudeInterface
+type: Model
+comment: >
+    Interface for a model responsible for simulating the measurements seen by
+    the CDGPS sensor during flight.
+
+args:
+    - satellite
+    - other
+
+params:
+    - name: "sensors.{satellite}.cdgps.model_range"
+      type: Integer
+      comment: >
+          Flag toggling whether or not being out of range will prevent the CDGPS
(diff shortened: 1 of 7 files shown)


kylekrol:
@shihaocao and whoever else is familiar with RTK, if we could verify exactly what they mean by baseline position for this PR that would be much appreciated!

kylekrol:
@shihaocao Per our brief discussion of RTK readouts, once I flip a minus sign I believe this model should be correct. We take the different between the two satellites in ECEF and return the vector pointing from the "other" satellite to "this" satellite.

src/psim/sensors/cdgps_no_attitude.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+Vector3 CdgpsNoAttitude::sensors_satellite_cdgps_dr() const {
+  auto const &truth_r_ecef = truth_satellite_orbit_r_ecef->get();
+  auto const &truth_other_r_ecef = truth_other_orbit_r_ecef->get();
+  auto const &error = Super::sensors_satellite_cdgps_dr_error.get();
+
+  /** If the sensor isn't active, the NaNs from the error field will make this
+   *  calculation results in NaNs as well as desired.
+   */
+  return (truth_other_r_ecef - truth_r_ecef) + error;
```
kylekrol:
```suggestion
  return (truth_r_ecef - truth_other_r_ecef) + error;
```

include/psim/sensors/cdgps_no_attitude.yml
```diff
@@ -0,0 +1,48 @@
...
+adds:
+    - name: "sensors.{satellite}.cdgps.active"
+      type: Lazy Integer
+      comment: >
+          Flag specifying whether or not the CDGPS sensor is currently active
+          and producing readings.
+    - name: "sensors.{satellite}.cdgps.dr"
+      type: Lazy Vector3
+      comment: >
+          Relative position of the other satellite reported by the CDGPS in ECEF.
```
kylekrol:
```suggestion
          Relative position this satellite to the other reported by the CDGPS in
          ECEF.
```

include/psim/sensors/cdgps_no_attitude.yml
```diff
@@ -0,0 +1,48 @@
...
+    - name: "sensors.{satellite}.cdgps.dr.error"
+      type: Lazy Vector3
+      comment: >
+          Error in the relative position of the other satellite by the CDGPS in
+          ECEF.
```
kylekrol:
```suggestion
          Error in the relative position of the this satellite to the other
          satellite by the CDGPS in ECEF.
```

shihaocao (approved):

PR #304, opened 2021-02-10
Fix Truth Hill Frame Model

kylekrol:
Long story short, I messed this up _a lot_ last time I implemented it. Will comment inline the fixes...

PR #305, opened 2021-02-10
Relative Orbit Square Root Kalman Filter

kylekrol:
Will be rebased once #304 is merged.


kylekrol:
Messed up my branching madness...

PR #307, opened 2021-02-10
Square Root Kalman Filter for Relative Orbit Estimation

### LINKED ISSUE #298: Add Relative Orbital State Estimation Function
We are already tracking work for an absolute orbit estimation function but don't have a ticket for an estimator that will be used once the two spacecraft are in CDGPS range - hence this ticket.

We'll most likely need to complete #141 along the way.

kylekrol:
Rebasing once #302 and #304 are merged.

Fixes #298.

### Summary of Changes

- Relative orbit estimator implementation as part of the `gnc` library.
- PSim relative orbit estimator model.
- PSim `RelativeOrbitEstimatorTest` simulation.

### Testing

Originally with a process noise value of `S = 1.0e-4 I` gave the following results:

image

which perform fairly well and although when the CDGPS goes out of range you can see the covariance bounds of the position estimate are very conservative - likely because the process noise for the position states is too large. Secondly, the covariance bounds on the seem too tight on the `z` velocity component. Taking this into account, the process noise values were updated.

The following results are shown below:

image

### DIFF
--- a/src/gnc/relative_orbit_estimate.cpp
+++ b/src/gnc/relative_orbit_estimate.cpp
@@ -1,0 +1,229 @@
+//
+// MIT License
+//
+// Copyright (c) 2020 Pathfinder for Autonomous Navigation (PAN)
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+//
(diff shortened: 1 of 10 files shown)


src/gnc/relative_orbit_estimate.cpp
```diff
@@ -0,0 +1,229 @@
...
+
+  return {
+    four - three * cnt,    zero, zero,     snt / n,               two * (one - cnt) / n,         zero,
+    six * (snt - nt),      one,  zero,     two * (cnt - one) / n, (four * snt - three * nt) / n, zero,
+    zero,                  zero, cnt,      zero,                  zero,                          snt / n,
+    three * n * snt,       zero, zero,     cnt,                   two * snt,                     zero,
+    six * n * (cnt - one), zero, zero,     -two * snt,            four * cnt - three,            zero,
+    zero,                  zero, -n * snt, zero,                  zero,                          cnt
+  };
+}
```
kylekrol, line 75: This function is incredibly important! Please verify the matrix elements against the linked reference. I've gone over them myself multiple times but we gotta make sure we get these right!

src/gnc/relative_orbit_estimate.cpp
```diff
@@ -0,0 +1,229 @@
...
+   *        = Q_hill_ecef * (dv_ecef + w_earth_ecef x dr_ecef - w_hill_ecef0 x dr_ecef0)
+   *        = Q_hill_ecef * (dv_ecef + (w_earth_ecef - w_hill_ecef) x dr_ecef)
+   */
+  r_hill = _Q_hill_ecef0 * (dr_ecef);
+  v_hill = _Q_hill_ecef0 * (dv_ecef + lin::cross(_w_earth_ecef - _w_hill_ecef0, dr_ecef));
+  _sqrtP = S;
+
+  _outputs();
+  _check_validity();
+}
```
kylekrol, line 210: Also please step through these frame conversions very carefully if possible!

src/gnc/relative_orbit_estimate.cpp
```diff
@@ -0,0 +1,229 @@
...
+void RelativeOrbitEstimate::_check_validity() {
+  if (lin::any(!lin::isfinite(_x)) || lin::any(!lin::isfinite(_dr_ecef)) ||
+      lin::any(!lin::isfinite(_dv_ecef)) || lin::any(!lin::isfinite(_sqrtP))) {
+    _valid = false;
+    _x = lin::nans<Vector<6>>();
+    _dr_ecef = lin::nans<Vector<3>>();
+    _dv_ecef = lin::nans<Vector<3>>();
+    _sqrtP = lin::nans<Matrix<6, 6>>();
+  }
+  _valid = true;
```
kylekrol:
```suggestion
  } else {
    _valid = true;
  }
```

shihaocao (approved):
matrix good

Issue #318, opened 2021-02-28
GRP Shadow Set Issues with the Attitude Estimator


kylekrol:
Sometimes the GRP attitude representation error just explodes with the attitude estimator. @stewartaslan mentioned that it may be because of a "shadow set" of GRP values. Not really sure what this means at the moment but negating the initial equation of the dynamics simulation solves the issue. To be clear, the filter actually doesn't perform well but it's GRP representation doesn't agree with the truth model - as in a negative sign is flipped but the estimated attitude is still correct.

The results can be currently replicated with the following command:

    python -m psim -s 1000 -p fc/attitude -ps 1 -c sensors/base,truth/base,truth/detumble -v AttitudeEstimatorTestGnc

on master with this commit 3d6c340ebf44ae317ec18bc3148afee739d7c6bf.

Some plots are given below. As you can see, the GRP error is horrid but actual degrees error and angular rate estimation error are still performing very well. The gyroscope bias estimate plots are not shown here but are behaving nominally.

![px](https://user-images.githubusercontent.com/33558436/109406688-2e391b00-7949-11eb-9a61-3c0aced9253a.png)
![qdeg](https://user-images.githubusercontent.com/33558436/109406694-33966580-7949-11eb-8e55-2893b9fd9024.png)
![w](https://user-images.githubusercontent.com/33558436/109406693-31cca200-7949-11eb-8e63-81e03cc85038.png)

kylekrol:
Marking this an noncritical given the results of #319.

PR #319, opened 2021-02-28
Potential Fix to the Shadow Set Issue

kylekrol:
Relates to #318.

### Summary of changes

- Negated a quaternion in the configuration files.

I just have a hunch our triad algorithm may always return a quaternion with a positive real component so...I just matched in and everything seems happy for now.

### Testing

The following plots were generated using the same command noted in the ticket:

    python -m psim -s 1000 -p fc/attitude -ps 1 -c sensors/base,truth/base,truth/detumble -v AttitudeEstimatorTestGnc

px
qdeg
w

### DIFF
--- a/config/parameters/truth/detumble.txt
+++ b/config/parameters/truth/detumble.txt
@@ -9,9 +9,9 @@ truth.t.ns  1700000000000
 
 # Leader spacecraft attitude and orbit initial conditions.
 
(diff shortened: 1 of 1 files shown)


shihaocao (approved):
Stewart has spoken.

Looks good to me if this is only a psim problem, and FSW still calculates the same

PR #322, opened 2021-03-07
Added standby initial conditions

### LINKED ISSUE #320: Standby Initial Conditions
We need initial conditions for HOOTL/HITL testing where the spacecraft is already detumbled to allow for faster testcase initialization.

Path forward: generate these new initial conditions by letting the spacecraft detumble for a period of time and then use the `--snapshot` utility -- note that some fields may need to be added to the snapshot utility!

kylekrol:
Closes #320.

### Summary of changes
- Added standby initial conditions. If using these in ptest, the satellite should pass right through detumble as it's angular rate should already be below the threshold.


kylekrol:
@shihaocao idk if we want to test this real quick in `ptest` prior to merging just to be sure!

shihaocao (approved):

PR #328, opened 2021-03-13
Feature/orbit controller

govindchari:
### Summary of Changes
- Does this by directly writing the desired impulse calculated by the orbit controller to the impulse field in the dynamics model when it is time to fire thrusters
- Enables the plotting of the relative distance and speed of the satellites as well as cumulative delta-V over time

### Testing
The plots of relative distance, relative speed, and cumulative delta-V are attached below. From these, it can be seen that the satellites rendezvous. 

dr
dv
deltaV

### DIFF
--- a/config/plots/truth/hill.yml
+++ b/config/plots/truth/hill.yml
@@ -1,0 +1,9 @@
+# Plots the distance between the satellites over time
+- type: Plot2D
+  x: truth.t.s
+  y: truth.leader.hill.dr.norm
+
+# Plots the relative velocity of the satellites over time
+- type: Plot2D
+  x: truth.t.s
+  y: truth.leader.hill.dv.norm
--- a/config/plots/truth/hill.yml
+++ b/config/plots/truth/hill.yml
@@ -7,3 +7,13 @@
 - type: Plot2D
   x: truth.t.s
   y: truth.leader.hill.dv.norm
+
+#Plots all three position coordinates in the Hill frame over time
+- type: Plot2D
+  x: truth.t.s
+  y: [truth.leader.hill.dr.x, truth.leader.hill.dr.y, truth.leader.hill.dr.z]
+
+#Plots all three velocity coordinates in the Hill frame over time
+- type: Plot2D
+  x: truth.t.s
+  y: [truth.leader.hill.dv.x, truth.leader.hill.dv.y, truth.leader.hill.dv.z]
--- a/include/psim/fc/orbit_controller.hpp
+++ b/include/psim/fc/orbit_controller.hpp
@@ -1,0 +1,56 @@
+//
+// MIT License
+//
+// Copyright (c) 2020 Pathfinder for Autonomous Navigation (PAN)
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+//
+
+/** @file psim/fc/orbit_controller.hpp
+ * @author Govind Chari
+ */
+
+#ifndef PSIM_FC_ORBIT_CONTROLLER_HPP_
+#define PSIM_FC_ORBIT_CONTROLLER_HPP_
+
+#include <psim/fc/orbit_controller.yml.hpp>
+
+#include <gnc/orbit_controller.hpp>
+
+namespace psim {
+
+class OrbitController : public OrbitControllerInterface<OrbitController> {
+ private:
+  typedef OrbitControllerInterface<OrbitController> Super;
+
+  gnc::OrbitControllerState _orbitController;
+
+  Integer last_firing = 0;
+
+ public:
+  using Super::OrbitControllerInterface;
+
+  OrbitController() = delete;
+  virtual ~OrbitController() = default;
+
+  virtual void step() override;
+};
+} // namespace psim
+
+#endif
--- a/include/psim/fc/orbit_controller.hpp
+++ b/include/psim/fc/orbit_controller.hpp
@@ -23,7 +23,7 @@
 //
 
 /** @file psim/fc/orbit_controller.hpp
- * @author Govind Chari
+ *  @author Govind Chari
  */
 
 #ifndef PSIM_FC_ORBIT_CONTROLLER_HPP_
@@ -39,9 +39,9 @@ class OrbitController : public OrbitControllerInterface<OrbitController> {
  private:
   typedef OrbitControllerInterface<OrbitController> Super;
 
-  gnc::OrbitControllerState _orbitController;
+  gnc::OrbitControllerState _orbit_controller;
 
-  psim::Integer last_firing = 0;
+  Integer last_firing = 0;
 
  public:
   using Super::OrbitControllerInterface;
--- a/include/psim/fc/orbit_controller.hpp
+++ b/include/psim/fc/orbit_controller.hpp
@@ -41,14 +41,14 @@ class OrbitController : public OrbitControllerInterface<OrbitController> {
 
   gnc::OrbitControllerState _orbitController;
 
-  Integer last_firing = 0;
+  psim::Integer last_firing = 0;
 
  public:
   using Super::OrbitControllerInterface;
 
   OrbitController() = delete;
   virtual ~OrbitController() = default;
-
+  virtual void add_fields(State &state) override;
   virtual void step() override;
 };
 } // namespace psim
--- a/include/psim/fc/orbit_controller.yml
+++ b/include/psim/fc/orbit_controller.yml
@@ -1,0 +1,27 @@
+
+name: OrbitControllerInterface
+type: Model
+comment: >
+    Interface for how the flight computer's orbit controller will interact with the rest of the simulation
+
+args:
+    - satellite
+    - other
+    - frame
+
+gets:
+    - name: "truth.t.s"
+      type: Real
+    - name: "truth.t.ns"
+      type: Integer
+    - name: "truth.{satellite}.orbit.r.ecef"
+      type: Vector3
+    - name: "truth.{satellite}.orbit.v.ecef"
+      type: Vector3
+    - name: "truth.{other}.orbit.r.ecef"
+      type: Vector3
+    - name: "truth.{other}.orbit.v.ecef"
+      type: Vector3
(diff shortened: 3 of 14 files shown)


src/psim/simulations/dual_orbit.cpp
```diff
@@ -47,6 +49,11 @@ DualOrbitGnc::DualOrbitGnc(
   add<SatelliteTruthNoAttitudeGnc>(randoms, config, "follower");
   add<HillFrameEci>(randoms, config, "leader", "follower");
   add<HillFrameEci>(randoms, config, "follower", "leader");
+  add<NormVector3>(randoms, config, "truth.leader.hill.dr");
+  add<NormVector3>(randoms, config, "truth.leader.hill.dv");
+
+
+
   // Sensors model
```
kylekrol: Let's clean up these extra blank lines here:
```suggestion
  add<NormVector3>(randoms, config, "truth.leader.hill.dv");
  // Sensors model
```

src/psim/simulations/orbit_controller_test.cpp
```diff
@@ -0,0 +1,42 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+//
+
+/** @file psim/simulations/orbit_controller_test.cpp
+ * @author Govind Chari
```
kylekrol:
```suggestion
 *  @author Govind Chari
```

src/psim/simulations/dual_orbit.cpp
```diff
@@ -34,6 +34,8 @@
 #include <psim/truth/hill_frame.hpp>
 #include <psim/truth/satellite_truth.hpp>
 #include <psim/truth/time.hpp>
+#include <psim/utilities/norm_vector3.hpp>
+
 
 namespace psim {
```
kylekrol:
```suggestion
#include <psim/utilities/norm_vector3.hpp>

namespace psim {
```

include/psim/simulations/orbit_controller_test.hpp
```diff
@@ -0,0 +1,48 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+//
+
+/** @file psim/simulations/orbit_controller_test.hpp
+ * @author Govind Chari
```
kylekrol:
```suggestion
 *  @author Govind Chari
```

include/psim/fc/orbit_controller.hpp
```diff
@@ -0,0 +1,56 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+//
+
+/** @file psim/fc/orbit_controller.hpp
+ * @author Govind Chari
```
kylekrol:
```suggestion
 *  @author Govind Chari
```

src/psim/fc/orbit_controller.cpp
```diff
@@ -0,0 +1,69 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+//
+
+/** @file psim/fc/orbit_controller.cpp
+ * @author Govind Chari
```
kylekrol:
```suggestion
 *  @author Govind Chari
```

src/psim/simulations/dual_attitude_orbit.cpp
```diff
@@ -34,6 +34,8 @@
 #include <psim/truth/hill_frame.hpp>
 #include <psim/truth/satellite_truth.hpp>
 #include <psim/truth/time.hpp>
+#include <psim/utilities/norm_vector3.hpp>
+
 
 namespace psim {
```
kylekrol:
```suggestion
#include <psim/utilities/norm_vector3.hpp>

namespace psim {
```

include/psim/fc/orbit_controller.yml
```diff
@@ -0,0 +1,33 @@
+
+name: OrbitControllerInterface
+type: Model
+comment: >
+    Interface for how the flight computer's orbit controller will interact with the rest of the simulation
```
kylekrol: If you don't have vertical rulers setup for your VSCode instance I'd recommend doing that [here](https://stackoverflow.com/questions/29968499/vertical-rulers-in-visual-studio-code)! We want to try and limit these files to an 80 character line limit especially for the comment sections so I'd go ahead and wrap the text at the top there.

include/psim/fc/orbit_controller.yml
```diff
@@ -0,0 +1,33 @@
...
+args:
+    - satellite
+    - other
+params:
+    - name: "truth.follower.m"
+      type: Real
+adds:
+    - name: "truth.follower.deltaV"
+      type: Real
+gets:
```
kylekrol:
```suggestion
args:
    - satellite
    - other

params:
    - name: "truth.{satellite}.m"
      type: Real

adds:
    - name: "truth.{satellite}.orbit.cummulative_dv"
      type: Real

gets:
```
Couple of comments.

First we have the two arguments `satellite` and `other` going which is awesome but we should use them to mangle the names of the parameters and fields we're adding here as well to keep the orbit controller model as portable as possible -- i.e. we could easily flip it to make the leader chasing the follower.

Second, field names are normally prefixed with a fixed string depending on what set of models is responsible for adding them to the simulation. In this case that means we should name anything being added by the orbit controller with `fc.**`. Also, just as a general naming convention, we don't have any fields that use camel case or capital letters in psim so I'd recommend renaming the `**.deltaV` field to something like `**.cummulative_dv` or something like that --  the `dx` to stand in for delta x is common throughout `psim`.

Lastly, we should add a comment field to added state fields. If someone if looking for a description of what's going on the interface file is going to be where they're looking.

include/psim/fc/orbit_controller.yml
```diff
@@ -0,0 +1,33 @@
...
+      type: Vector3
+    - name: "truth.{other}.orbit.r.ecef"
+      type: Vector3
+    - name: "truth.{other}.orbit.v.ecef"
+      type: Vector3
+    - name: "truth.{satellite}.orbit.J.ecef"
+      type: Writable Vector3
+
+
+
```
kylekrol, line 33:
```suggestion
      type: Writable Vector3

```

kylekrol: We should try and cleanup the lines at the end here too.

include/psim/fc/orbit_controller.hpp
```diff
@@ -0,0 +1,56 @@
...
+
+namespace psim {
+
+class OrbitController : public OrbitControllerInterface<OrbitController> {
+ private:
+  typedef OrbitControllerInterface<OrbitController> Super;
+
+  gnc::OrbitControllerState _orbitController;
+
+  psim::Integer last_firing = 0;
```
kylekrol: Because you're already in the `psim` namespace here you don't need to be this specific with the type information!
```suggestion
  Integer last_firing = 0;
```

include/psim/fc/orbit_controller.hpp
```diff
@@ -0,0 +1,56 @@
...
+
+#include <gnc/orbit_controller.hpp>
+
+namespace psim {
+
+class OrbitController : public OrbitControllerInterface<OrbitController> {
+ private:
+  typedef OrbitControllerInterface<OrbitController> Super;
+
+  gnc::OrbitControllerState _orbitController;
```
kylekrol: We want to try and stick to snake case for member variables and PascalCase for class/type names in general:
```suggestion
  gnc::OrbitControllerState _orbit_controller;
```

src/psim/fc/orbit_controller.cpp
```diff
@@ -0,0 +1,69 @@
...
+ */
+
+#include <lin/core.hpp>
+#include <lin/generators.hpp>
+#include <lin/math.hpp>
+#include <lin/queries.hpp>
+#include <psim/fc/orbit_controller.hpp>
```
kylekrol: If we're implementing something outlined in a header file we generally want to include that file along at the top to be explicit about what is getting implemented:
```suggestion
#include <psim/fc/orbit_controller.hpp>

#include <lin/core.hpp>
#include <lin/generators.hpp>
#include <lin/math.hpp>
#include <lin/queries.hpp>
```

src/psim/fc/orbit_controller.cpp
```diff
@@ -0,0 +1,69 @@
...
+    last_firing = t_ns;
+    gnc::OrbitControllerData data;
+    data.t = t;
+    data.r_ecef = r_ecef;
+    data.v_ecef = v_ecef;
+    data.dr_ecef = (other_r_ecef - r_ecef);
+    data.dv_ecef = (other_v_ecef - v_ecef);
+    gnc::OrbitActuation actuation;
+    gnc::control_orbit(_orbitController, data, actuation);
+    J_ecef = actuation.J_ecef;
```
kylekrol: There is potential bug here if the orbit controllers returns `nans` for the recommended impulse -- this i don't think ever happens but some controllers/estimators do it so we want to be sure the impulse contains finite values before passing it on to the rest of the simulation. In general, list can be done with:

    #include <lin/core.hpp>
    #include <lin/math.hpp>
    #include <lin/queries.hpp>

    ...

    if (lin::all(lin::isfinite(...))) {
      ...
    }

If the recommended impulse isn't finite we can just zero out the impulse with `lin::zeros<Vector3>()`.

kylekrol (changes requested):
Overall, everything looks great - most of the comments are style/syntax oriented.

config/plots/truth/hill.yml
```diff
@@ -0,0 +1,9 @@
+# Plots the distance between the satellites over time
+- type: Plot2D
+  x: truth.t.s
+  y: truth.leader.hill.dr.norm
+
+# Plots the relative velocity of the satellites over time
+- type: Plot2D
+  x: truth.t.s
+  y: truth.leader.hill.dv.norm
```
kylekrol: We also probably want to add plots of the x, y, and z hill frame position and velocity components over time? Commonly with this controller the `z` convergence is far slower because matching orbital planes is a little more challenging to control -- adding these plots would make that behavior apparent if rendezvous times were seeming long.

kylekrol (approved):
It looks good to me -- although we still have extra blank lines at the end of the orbit controller YAML file that we should clean up!

kylekrol:
Also @Govind-Chari it seems like you aren't signed in on github? I would look into `git config --global user.email` and `git config --global user.email`.

PR #330, opened 2021-03-20
Created orbit controller unit test

govindchari:
# Orbit Controller Unit Test

### Summary of changes
Created a unit test for orbit controller. Throws error if spacecrafts fail to rendezvous in 10 days. Here, rendezvous is defined as <0.5m.

### Testing
Currently, rendezvous occurs in about 9.5 days, so 10 days was chosen as the threshold for regression. Below are plots of the current performance of the simulation. For reference, 821,000 seconds is ~9.5 days.
dr_vec
dr_norm
dr_norm_zoomed

### DIFF
--- a/python/test/test_orbit_controller.py
+++ b/python/test/test_orbit_controller.py
@@ -1,0 +1,24 @@
+from psim import Configuration, sims, Simulation
+
+import lin
+import pytest
+
+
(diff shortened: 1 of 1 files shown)


kylekrol (approved):
Assuming CI is still happy it looks good to me.

PR #341, opened 2021-04-02
Estimator controller integration

govindchari:
# Integrated OrbitEstimator and RelativeOrbitEstimator into OrbitController

### Summary of changes
- OrbitController uses estimated states instead of truth states

### Testing
No testing is needed to see that the integration of the estimators and controllers occurs


src/psim/fc/orbit_controller.cpp
```diff
@@ -45,22 +45,51 @@ void OrbitController::step() {
...
+    dr_ecef_smoothed=dr_ecef;
+  } else {
+    dr_ecef_smoothed=alpha*dr_ecef+(1.0-alpha)*prev_dr_ecef;
+  }
+
+  if(!lin::all(lin::isfinite(prev_dv_ecef))){
+    dv_ecef_smoothed=dv_ecef;
+  } else {
+    dv_ecef_smoothed=alpha*dv_ecef+(1.0-alpha)*prev_dv_ecef;
+  }
```
kylekrol: I don't think we're ever actually doing any smoothing here? Like we never assign back to `prev_d*_ecef` so it's the same as reading data straight from the estimator.

include/psim/fc/orbit_controller.hpp
```diff
@@ -42,6 +43,10 @@ class OrbitController : public OrbitControllerInterface<OrbitController> {
   gnc::OrbitControllerState _orbit_controller;
 
   Integer last_firing = 0;
+  Vector3 prev_dr_ecef=lin::nans<Vector3>();
+  Vector3 prev_dv_ecef=lin::nans<Vector3>();
+  Integer alpha=0.5;
+
```
kylekrol: Also could just run clang format through here to cleanup the extra blank lines, equal sign spacing, etc.

config/parameters/truth/final_approach.txt
```diff
@@ -0,0 +1,26 @@
...
+
+# Time initial conditions.
+
+truth.t.ns 556410000000000
+
+# Leader spacecraft attitude and orbit initial conditions.
+
+truth.leader.attitude.q.body_eci  0.173301   0.971103 0.029768   0.161367
+truth.leader.attitude.w           0.00359105 0.137562 0.00386897
+truth.leader.wheels.w             0.0        0.0      0.0
```
kylekrol, line 14: It seems like these attitude initial conditions were pulled form a standby case? I think that's the way to go but just wanted to double check!

kylekrol (changes requested):
Unless I'm totally missing something, it doesn't seem like the smoothing was being done at all?

config/parameters/sensors/base.txt
```diff
@@ -21,7 +21,7 @@ sensors.leader.gps.v.sigma  5.0 5.0 5.0
 
 sensors.leader.cdgps.model_range  1
 sensors.leader.cdgps.range        2000.0
-sensors.leader.cdgps.dr.sigma     0.1    0.1 0.1
+sensors.leader.cdgps.dr.sigma     0.01  0.01 0.01
```
kylekrol, line 26: Just want to double check here, do we actually have a source for this? If not, we should open up a ticket in `psim` as well as flight software because the sensor noise covariances in both repositories are set off this number.

src/gnc/orbit_controller.cpp
```diff
@@ -39,26 +39,26 @@
...
-static constexpr double mass = 3.7;             // Mass of spacecraft            (kg)
-static constexpr double J_min = 0;              // Minimum impulse               (N s)
-static constexpr double J_max = 2.5e-2;         // Maximum impulse               (N s)
-static constexpr double max_dv = J_max / mass;  // Max velocity change           (m/s)
-static constexpr double min_dv = J_min / mass;  // Min velocity change           (m/s)
+static constexpr double mass = 3.7;     // Mass of spacecraft            (kg)
+static constexpr double J_min = 0;      // Minimum impulse               (N s)
+static constexpr double J_max = 2.5e-2; // Maximum impulse               (N s)
+static constexpr double max_dv = J_max / mass; // Max velocity change (m/s)
+static constexpr double min_dv = J_min / mass; // Min velocity change (m/s)
```
kylekrol: What are the functional changes in this file? If it's not that many we should revert this file to what's on master, just apply those changes, and leave the diff as minimal as possible so we don't obliterate the git blame here.

kylekrol (changes requested):
Other than the formatting of the `gnc` orbit controller implementation, everything looks great. If there aren't functional changes we shouldn't throw all the formatting changes in with this PR as it'll just clutter the diff.

In general, the `gnc` library was all written prior to using clang format so we can let those files be format wise to avoid bloated diffs. Ping me on slack if there are functional changes that I missed but it didn't seem like there were any.

kylekrol (approved):

PR #344, opened 2021-04-14
Variable gain

govindchari:
# Variable Fire time and Gains

### Summary of changes
- Gains and fire time transition when satellites get into near-field

### Testing
Plots from long-duration  simulation are attached

Screen Shot 2021-04-13 at 8 53 14 PM
Screen Shot 2021-04-13 at 8 53 37 PM
Screen Shot 2021-04-13 at 8 54 03 PM


kylekrol (approved):
Lgtm!

PR #349, opened 2021-05-08
NF Firing ICs

govindchari:
# Quick Firing ICs

### Summary of changes
- Creates a near and far-field initial conditions file with a firing node within a minute


kylekrol (approved):
Looks good to me! We want to add mission cases for near field firing and far field firing too?

shihaocao:
@govindchari do we still need these

govindchari:
> @govindchari do we still need these

no

PR #350, opened 2021-09-27
Allows PSim to run without sun vectors 

govindchari:
# PSim No Sun Vector Update

### Summary of changes
- In PR #795 in FSW the functionality to turn off sun vectors was added, but the fc model in psim was not updated
- This PR updates that model

### DIFF
--- a/src/psim/fc/attitude_estimator.cpp
+++ b/src/psim/fc/attitude_estimator.cpp
@@ -76,8 +76,13 @@ void AttitudeEstimator::step() {
   }
   // Attempt to reset the current estimate if it isn't valid.
   else {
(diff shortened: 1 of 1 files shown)


govindchari:
Screen Shot 2021-09-27 at 1 00 56 AM

Screen Shot 2021-09-27 at 1 00 10 AM

The two above graphs are two simulations run with the changes. The top graph is attitude estimate error when the sun sensors are active and the bottom is attitude estimate error without sun vectors. You can see that this is the case since the bottom graph has worse performance which is intuitively correct.

shihaocao (approved):
Good stuff good stuff.

PR #354, opened 2022-01-07
Implemented Noise Radius and Config on Thruster Impulse

shihaocao:
# Implemented Noise Radius and Config on Thruster Impulse

This PR adds a spherical gaussian noise on the thruster impulse. The radius of the noise is specified in a config file.

TODO: do the math to specifiy a good noise radius.

### DIFF
--- a/config/parameters/fc/base.txt
+++ b/config/parameters/fc/base.txt
@@ -7,5 +7,10 @@ fc.follower.fire_time_far     1800
 fc.follower.fire_time_near    300
 
 # Ideally these would be in an actuators folder...
-fc.leader.thruster.noise_radius 1.0
-fc.follower.thruster.noise_radius 1.0
+# https://docs.google.com/spreadsheets/d/1PmbRiLaxWfPm8o3WJnG9EudwjCTLcfDGJe1scMyEqao/edit?usp=sharing
+# Best Case 0.0002828427125
+# Expected Case 0.0004242640687
+# Worst Case (Conservative) 0.0008485281374
+
+fc.leader.thruster.noise_sigma 0.0004242640687
+fc.follower.thruster.noise_sigma 0.0004242640687
--- a/config/parameters/fc/base.txt
+++ b/config/parameters/fc/base.txt
@@ -5,3 +5,7 @@ fc.leader.fire_time_near   300
 
 fc.follower.fire_time_far     1800
 fc.follower.fire_time_near    300
+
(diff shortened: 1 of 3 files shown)


config/parameters/fc/base.txt
```diff
@@ -5,3 +5,7 @@ fc.leader.fire_time_near   300
 
 fc.follower.fire_time_far     1800
 fc.follower.fire_time_near    300
+
+# Ideally these would be in an actuators folder...
+fc.leader.thruster.noise_radius 1.0
```
govindchari: renaming this thruster.noise.1sig would be more accurate just so we know it is 1sig

govindchari:
Other than the name of the noise config parameter this pr lgtm

govindchari:
ImpulseVsOnTime_wResiduals

Make the noise 1e-3 Ns 3-sigma. This is conservative as can be seen from the above plot residuals. The assumption that we use is that by being conservative, we do not have to explicitly account for the fact that there are four valves rather that three valves aligned with the coordinate axes of the satellite.

config/parameters/fc/base.txt
shihaocao: will rename!

shihaocao:
I just fixed another bug I think, I normalized the vector before multiplying by the sigma. Otherwise, I think we could potentially be adding more variance than intended?

Unsure if this is how the math works out lol...

config/parameters/fc/base.txt
shihaocao: i think _sigma is sufficient

src/psim/fc/orbit_controller.cpp
```diff
@@ -114,7 +116,18 @@ void OrbitController::step() {
     gnc::control_orbit(_orbit_controller, data, actuation);
 
     if (lin::all(lin::isfinite(actuation.J_ecef))) {
-      J_ecef = actuation.J_ecef;
+      //get random noise
+      Vector3 random_noise = lin::gaussians<Vector3>(_randoms);
+
+      // normalize the random noise
+      random_noise = random_noise / lin::norm(random_noise);
```
govindchari: don't normalize

shihaocao: Lol yeah this is dumb, forgot the extra gaussian anyway

govindchari (approved):
lgtm

